PR #86, opened 2026-04-06
feat: warn with observation diff when reset reproducibility fails

Shailesh-05:
image

Screenshot taken with CUBE_DEMO_RESET_REPRO=1 so reset reproducibility fails on purpose for demo; uv run cube test counter-cube without that variable still passes reset repro.

**How to Test:**
`uv run pytest tests/test_testing.py tests/test_cli.py`
Optional — see the yellow panel locally:
`CUBE_DEMO_RESET_REPRO=1 uv run cube test counter-cube`


amanjaiswal73892 (changes requested):
LGTM! Let's add the changes we discussed and one final pass by @NicolasAG  and we can merge.

Shailesh-05:
> image
> Screenshot taken with CUBE_DEMO_RESET_REPRO=1 so reset reproducibility fails on purpose for demo; uv run cube test counter-cube without that variable still passes reset repro.
> 
> **How to Test:** `uv run pytest tests/test_testing.py tests/test_cli.py` Optional — see the yellow panel locally: `CUBE_DEMO_RESET_REPRO=1 uv run cube test counter-cube`

@NicolasAG  @amanjaiswal73892 LGTM.

NicolasAG:
### Code review

Found 2 issues:

1. CI mode prints `reset_diff` without truncation — the dashboard path applies `_RESET_DIFF_DISPLAY_MAX = 24_000` chars, but the CI branch prints the raw diff unconditionally. For cubes with large observations (e.g. base64 screenshots), this can flood CI logs with hundreds of kilobytes.

**Fix:** apply the same cap in the CI path before printing:
\`\`\`python
if reset_diff.strip():
    print(reset_diff[:_RESET_DIFF_DISPLAY_MAX].rstrip("\n"))
\`\`\`

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/cli.py#L127-L129

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/cli.py#L573-L576

2. `_DEMO_RESET_REPRO_DIFF` uses unified diff syntax (`---/+++/@@`) but `format_observation_unified_diff` returns key-path format (`Observation differences\n\nsome.key\n  first: X\n  second: Y`). The stated purpose of `--demo-reset-repro` is to preview the real error UI, but the demo renders a completely different format than an actual failure would.

**Fix:** replace `_DEMO_RESET_REPRO_DIFF` with output actually produced by `format_observation_unified_diff`, e.g.:
\`\`\`python
from cube.core import Observation
_DEMO_RESET_REPRO_DIFF = format_observation_unified_diff(
    Observation.from_text("Counter starts at 0. Use 'increment' action to reach 3."),
    Observation.from_text("Counter starts at 0. Use 'increment' action to reach 3. (demo token: a1b2c3d4e5f6)"),
)
\`\`\`
Also rename `format_observation_unified_diff` → `format_observation_diff` to match what it actually produces.

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/cli.py#L134-L141

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/testing.py#L255-L259

examples/counter-cube/src/counter_cube/task.py
```diff
@@ -31,7 +33,12 @@ def target(self) -> int:
     def reset(self) -> tuple[Observation, dict[str, Any]]:
         """Reset tool state and return the opening observation."""
         self.tool.reset()
-        obs = Observation.from_text(f"Counter starts at 0. Use 'increment' action to reach {self.target}.")
+        text = f"Counter starts at 0. Use 'increment' action to reach {self.target}."
+        # Opt-in only: two fresh Task instances then get different first observations, so
+        # `cube test` can surface the reset-reproducibility panel for screenshots / demos.
+        if os.environ.get("CUBE_RESET_REPRO_SNAPSHOT_DEMO"):
+            text = f"{text} (demo token: {uuid.uuid4().hex[:12]})"
+        obs = Observation.from_text(text)
```
NicolasAG: This feels like a hack that was coded just to take a screenshot at some point... Can we safely revert this change?

NicolasAG:
### Additional code review notes

**General architectural note:** error messages (including reset-reproducibility failures) should use plain `print()` rather than Rich markup. Rich formatting should be reserved for the final results dashboard/table. This avoids the markup-escaping issues below and keeps the error path simpler.

---

5 additional issues:

1. **Leftover screenshot hack should be removed** — `CUBE_RESET_REPRO_SNAPSHOT_DEMO` was added to the counter-cube example solely to inject a non-deterministic token so the author could capture a screenshot of the reset-reproducibility panel. The comment confirms this: `"…so cube test can surface the reset-reproducibility panel for screenshots / demos."` This should not stay in the example.

**Fix:** revert `examples/counter-cube/src/counter_cube/task.py` to its state before this PR — remove the `if os.environ.get("CUBE_RESET_REPRO_SNAPSHOT_DEMO")` block and the `import os` / `import uuid` lines. If a demo is still needed, use `--demo-reset-repro` / `CUBE_DEMO_RESET_REPRO=1` which already bypasses the real check in the CLI without touching any cube code.

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/examples/counter-cube/src/counter_cube/task.py#L36-L42

2. **`MarkupError` crash when exception message contains bracket characters** — `reset_msg` (derived from `str(e)`) is interpolated directly into a Rich markup string without escaping. If any exception raised inside `reset()` contains characters like `[/dim]` or `[red]`, `Text.from_markup` raises `rich.errors.MarkupError` and crashes the CLI before showing any output.

**Fix:** per the architectural note above, switch this path to plain `print()`. If Rich is kept, escape the user-derived portion:
\`\`\`python
Text.assemble(Text(intro), Text.from_markup(" [dim]A mismatch is not always a bug.[/dim]"))
\`\`\`

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/cli.py#L155-L160

3. **Misleading prefix on pre-instantiation errors** — `"(first task, two fresh Task instances):"` is prepended to every reset-reproducibility error, but `check_reset_reproducibility` has early-return paths (`"no get_debug_benchmark"`, `"no debug task configs"`) that fire before any Task is instantiated. The prefix is factually wrong for those cases and misdirects the user.

**Fix:** only add the prefix when `reset_msg` describes an actual observation mismatch, or pass a flag from `check_reset_reproducibility` to distinguish configuration errors from comparison failures. The simplest approach is to include the contextual info inside the error strings returned by the early-return paths themselves.

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/cli.py#L154-L160

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/cli.py#L569-L574

4. **`--demo-reset-repro` + `--ci` produces contradictory output** — when both flags are active on a clean run, the demo block appends `test_reset_reproducibility` to `compliance_failed` (so CI output prints `FAIL  test_reset_reproducibility`) but then, because `failures` is still empty, the process exits 0 and prints `PASSED`. A CI log parser that scans for `FAIL` will report a failure while the exit code says success.

**Fix:** guard the demo flag against CI mode:
\`\`\`python
if demo_reset_repro and not ci_mode and not failures and reset_ok:
\`\`\`
Or raise an error early if both `--ci` and `--demo-reset-repro` are passed together.

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/cli.py#L529-L535

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/cli.py#L581-L584

5. **Non-Pydantic observation equality regression** — the fallback for observations without `model_dump` changed from `str(obs)` to the raw object. For any observation class without a custom `__eq__`, Python compares by identity: two distinct instances with identical content will always be unequal, causing `check_reset_reproducibility` to report a spurious failure.

**Fix:** revert the fallback to `str(obs)`:
\`\`\`python
dump1 = obs1.model_dump() if hasattr(obs1, "model_dump") else str(obs1)
dump2 = obs2.model_dump() if hasattr(obs2, "model_dump") else str(obs2)
\`\`\`

https://github.com/The-AI-Alliance/cube-standard/blob/193db622789678d06751c75a064842de4efe571c/src/cube/testing.py#L294-L300

examples/counter-cube/src/counter_cube/task.py
Shailesh-05: You’re right, that was only there so two fresh Task instances could get different first observations for screenshots. It isn’t needed for real behavior. Thanks for catching it!

amanjaiswal73892:
### Minor nits (non-blocking)

Three small things I noticed that aren't covered in the existing review thread:

**1. `_RESET_DIFF_DISPLAY_MAX` placement** ([cli.py L129](https://github.com/The-AI-Alliance/cube-standard/blob/5542953034b066dbdb7e659ab5ae10c943313992/src/cube/cli.py#L129))

This constant is defined in `cli.py` but logically belongs alongside the diff helpers in `testing.py`. Both the CI path (`_emit_reset_repro_plain`) and the dashboard path read it from `cli.py`, so a future contributor adding a new diff consumer in `testing.py` is likely to miss it or set a different cap. Low priority — not a blocker.

---

**2. `call_count == 4` test assertions are tightly coupled to throughput strategy** ([test_cli.py L225–226](https://github.com/The-AI-Alliance/cube-standard/blob/5542953034b066dbdb7e659ab5ae10c943313992/tests/test_cli.py#L225-L226))

```python
assert mock_suite.call_count == 4
assert [c.kwargs.get("workers") for c in mock_suite.call_args_list] == [1, 1, 2, 4]
```

This pins the exact call sequence (1 compliance + 1 warm-up + workers 2 + workers 4). If the strategy changes (warm-up removed, worker counts adjusted) these will fail for the wrong reason. Consider asserting on behavior instead — e.g. the compliance call uses `workers=1` and at least one call uses `workers > 1`.

---

**3. `format_observation_diff` interface ambiguity** ([testing.py L248](https://github.com/The-AI-Alliance/cube-standard/blob/5542953034b066dbdb7e659ab5ae10c943313992/src/cube/testing.py#L248), [L279–282](https://github.com/The-AI-Alliance/cube-standard/blob/5542953034b066dbdb7e659ab5ae10c943313992/src/cube/testing.py#L279-L282))

`check_reset_reproducibility` already calls `model_dump()` on both observations before passing them to `format_observation_diff`, which also calls `model_dump()` internally. For dict inputs the double-dump is harmless, but it's unclear from the signature whether the function expects raw Pydantic objects or pre-dumped dicts. A short docstring note would prevent confusion if this function gets reused elsewhere.

amanjaiswal73892 (approved):
LGTM! Thanks Shailesh, We can merge after addressing prefix error (see number 3 from Nicolas's comment).

Shailesh-05:
> LGTM! Thanks Shailesh, We can merge after addressing prefix error (see number 3 from Nicolas's comment).

Sounds good @amanjaiswal73892 , prefix error has been resolved. LGTM Shall I merge?

PR #89, opened 2026-04-08
feat: NeMo Gym integration

recursix:
## Summary

- Adds `CubeResourcesServer` in `cube.integrations.nemogym` that wraps any CUBE `Benchmark` as an HTTP server compatible with NeMo Gym's resource server protocol (`/seed_session`, `/step`, `/verify`, `/close`)
- Adds `Action.from_openai_tool_call()` to parse both Chat Completions and Responses API tool call formats into CUBE `Action` objects
- No NeMo Gym Python dependency required -- pure HTTP contract

This lets NeMo Gym's `CubeAgent` orchestrate multi-step rollouts against any CUBE benchmark without importing CUBE internals. Related: NVIDIA-NeMo/Gym#1006

## Test plan

- [x] 4 unit tests for `Action.from_openai_tool_call()` (flat, nested, dict args, call_id)
- [x] 9 integration tests for `CubeResourcesServer` (health, seed, step, verify, close, full episode, concurrent sessions, error cases)
- [x] All 77 existing tests still pass
- [x] Ruff lint clean

### DIFF
--- a/src/cube/integrations/nemogym.py
+++ b/src/cube/integrations/nemogym.py
@@ -1,0 +1,209 @@
+"""
+NeMo Gym integration for CUBE.
+
+Provides a CubeResourcesServer that wraps any CUBE Benchmark as an HTTP server
+compatible with NeMo Gym's resource server protocol. NeMo Gym's CubeAgent calls
+these endpoints over HTTP — no NeMo Gym Python dependency required.
+
+Endpoints:
+    POST /seed_session  — pick a task, reset it, return initial observation + tools
+    POST /step          — execute an action, return next observation + reward + done
+    POST /verify        — evaluate the current task state, return reward
+    POST /close         — close a task and free resources
+
+Usage:
+    from cube.integrations.nemogym import CubeResourcesServer
+
+    server = CubeResourcesServer(benchmark=my_benchmark)
+    server.run(host="0.0.0.0", port=8080)
+
+Or programmatically:
+    app = server.make_app()  # returns a FastAPI instance
+"""
+
+import atexit
+import logging
+import uuid
+from contextlib import asynccontextmanager
+
+import uvicorn
+from fastapi import FastAPI, HTTPException
+from pydantic import BaseModel
+
+from cube.benchmark import Benchmark
+from cube.core import Action
+from cube.task import Task, TaskConfig
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Request / Response schemas
+# ---------------------------------------------------------------------------
+
+
+class SeedSessionRequest(BaseModel):
+    task_idx: int
+
+
+class SeedSessionResponse(BaseModel):
+    env_id: str
+    observation: list[dict]
+    tools: list[dict]
+
+
+class StepRequest(BaseModel):
+    env_id: str
+    action: dict
+
+
+class StepResponse(BaseModel):
+    observation: list[dict]
+    reward: float
+    done: bool
+
+
+class VerifyRequest(BaseModel):
+    env_id: str
+
+
+class VerifyResponse(BaseModel):
+    reward: float
+    reward_info: dict = {}
+
+
+class CloseRequest(BaseModel):
+    env_id: str
+
+
+# ---------------------------------------------------------------------------
+# Server
+# ---------------------------------------------------------------------------
+
+
+class CubeResourcesServer:
+    """Wraps a CUBE Benchmark as a NeMo Gym-compatible HTTP resource server.
+
+    Manages multiple concurrent sessions (env_id → Task), each corresponding
+    to one episode. The benchmark must be set up before constructing this server.
+
+    Args:
+        benchmark: A CUBE Benchmark instance (already set up via benchmark.setup()).
+    """
+
+    def __init__(self, benchmark: Benchmark) -> None:
+        self.benchmark = benchmark
+        self._task_configs: list[TaskConfig] = list(benchmark.get_task_configs())
+        self._sessions: dict[str, Task] = {}
+        self._last_obs: dict[str, object] = {}  # env_id → last Observation for verify
+
+        if not self._task_configs:
+            raise ValueError(f"Benchmark '{benchmark.name}' has no task configs")
+
+        logger.info("CubeResourcesServer loaded %d tasks from '%s'", len(self._task_configs), benchmark.name)
+
+    # -- Lifecycle -----------------------------------------------------------
+
+    def _close_all(self) -> None:
+        """Best-effort cleanup of all open sessions."""
+        for env_id, task in list(self._sessions.items()):
+            try:
+                task.close()
+                logger.debug("Closed session %s", env_id)
+            except Exception:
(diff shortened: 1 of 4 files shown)


src/cube/integrations/nemogym.py
```diff
@@ -0,0 +1,211 @@
...
+    Manages multiple concurrent sessions (env_id → Task), each corresponding
+    to one episode. The benchmark must be set up before constructing this server.
+
+    Args:
+        benchmark: A CUBE Benchmark instance (already set up via benchmark.setup()).
+    """
+
+    def __init__(self, benchmark: Benchmark) -> None:
+        self.benchmark = benchmark
+        self._task_configs: list[TaskConfig] = list(benchmark.get_task_configs())
```
NicolasAG, line 104: some benchmarks will need to call `.install()` and `.setup()` before calling `.get_task_configs()` -- to be confirmed after uniformization if this will still hold though.

src/cube/integrations/nemogym.py
```diff
@@ -0,0 +1,211 @@
...
+        if body.task_idx < 0 or body.task_idx >= len(self._task_configs):
+            raise HTTPException(
+                status_code=400, detail=f"task_idx {body.task_idx} out of range [0, {len(self._task_configs)})"
+            )
+
+        task_config = self._task_configs[body.task_idx]
+        task = task_config.make(
+            runtime_context=self.benchmark._runtime_context,
+            container_backend=self.benchmark.container_backend,
+        )
```
NicolasAG, line 141: this definitely needs a `benchmark.install() && benchmark.setup()` call before. It will still be the case after uniformization.

src/cube/integrations/nemogym.py
```diff
@@ -0,0 +1,211 @@
...
+    def verify(self, body: VerifyRequest) -> VerifyResponse:
+        task = self._sessions.get(body.env_id)
+        if task is None:
+            raise HTTPException(status_code=404, detail=f"Unknown env_id: {body.env_id}")
+
+        obs = self._last_obs.get(body.env_id)
+        if obs is None:
+            raise HTTPException(status_code=400, detail=f"No observation recorded for env_id: {body.env_id}")
+
+        reward, info = task.evaluate(obs)
```
NicolasAG, line 187: `obs` will actually be optional in the future. Many cubes don't use it anyway.
I have this coded already in the `json-rpc` branch: https://github.com/The-AI-Alliance/cube-standard/blob/json-rpc/src/cube/task.py#L263

NicolasAG:
### Code review

Small issues worth considering before merging:

1. **Resource leak if `task.reset()` raises** — `task_config.make()` may allocate containers/connections, but if `reset()` throws, the task is never added to `_sessions` and `_close_all()` won't reach it.

https://github.com/The-AI-Alliance/cube-standard/blob/85c9db80057f0da7f044d764a60a50de2d4818ab/src/cube/integrations/nemogym.py#L126-L136

Suggested fix: wrap in `try/except` and call `task.close()` on failure before re-raising.

2. **`arguments: null` crash in `from_openai_tool_call`** — `.get("arguments", "{}")` only uses the default when the key is absent. If a caller sends `"arguments": null` (valid JSON), `get()` returns `None`, the `isinstance(raw_args, str)` check is `False`, and Pydantic validation fails on `arguments=None` instead of falling back to `{}`.

https://github.com/The-AI-Alliance/cube-standard/blob/85c9db80057f0da7f044d764a60a50de2d4818ab/src/cube/core.py#L155-L160

Suggested fix: `raw_args = tool_call.get("arguments") or "{}"` (or an explicit `None` check).

3. **`/verify` evaluates with post-processed obs** — `step()` stores `env_output.obs` in `_last_obs`, but that observation has already been through `task.obs_postprocess()`. Internally, `task.step()` calls `evaluate()` on the raw obs *before* post-processing. Calling `task.evaluate(post_processed_obs)` in `/verify` is inconsistent with that contract and can produce different rewards for tasks whose `obs_postprocess` mutates the observation.

https://github.com/The-AI-Alliance/cube-standard/blob/85c9db80057f0da7f044d764a60a50de2d4818ab/src/cube/integrations/nemogym.py#L151-L153

https://github.com/The-AI-Alliance/cube-standard/blob/85c9db80057f0da7f044d764a60a50de2d4818ab/src/cube/integrations/nemogym.py#L168-L170

Suggested fix: store the raw obs before post-processing, or call `task.step()` and let it handle evaluation internally.

4. **`EnvironmentOutput.error` is silently dropped** — when a step fails with a `StepError`, `task.step()` sets `done=True` and populates `env_output.error`. The `StepResponse` doesn't include an `error` field, so clients can't distinguish a failed step from a successful terminal one.

https://github.com/The-AI-Alliance/cube-standard/blob/85c9db80057f0da7f044d764a60a50de2d4818ab/src/cube/integrations/nemogym.py#L154-L158

Suggested fix: add `error: str | None = None` to `StepResponse` and populate it from `env_output.error`.

5. **`_close_all()` called twice on shutdown + atexit accumulates across `make_app()` calls** — the `lifespan` context manager already calls `_close_all()` on shutdown; `atexit.register(self._close_all)` adds a second call. Since `_close_all()` clears `_sessions` on its first run, the `atexit` call is a no-op on clean shutdown — which defeats its purpose as a safety net. Also, each call to `make_app()` (e.g. in tests) registers a new `atexit` entry without deduplication.

https://github.com/The-AI-Alliance/cube-standard/blob/85c9db80057f0da7f044d764a60a50de2d4818ab/src/cube/integrations/nemogym.py#L185-L194

Suggested fix: remove `atexit.register`; `lifespan` alone is sufficient (consistent with `server.py`).

6. **No endpoint to enumerate available tasks** — `/seed_session` takes a `task_idx: int`, but there is no endpoint to list tasks or their metadata. Callers must hard-code indices or probe blindly. The existing `server.py` solves this with a `/cube/tasks` endpoint returning `list[TaskMetadata]`.

https://github.com/The-AI-Alliance/cube-standard/blob/85c9db80057f0da7f044d764a60a50de2d4818ab/src/cube/integrations/nemogym.py#L202-L204

Suggested fix: add a `GET /tasks` endpoint returning `[{"idx": i, "id": tc.task_id, ...} for i, tc in enumerate(self._task_configs)]`.

recursix:
Thanks Nicolas, all good catches.

Fixed in the latest push:

1. **Resource leak** - wrapped `task.reset()` in try/except, calls `task.close()` on failure before re-raising.
2. **`arguments: null`** - changed to `or "{}"` so explicit null falls back to empty dict.
3. **Post-processed obs in `/verify`** - this can't be cleanly fixed without a Task API change (EnvironmentOutput only carries the post-processed obs). Added a comment documenting the known limitation. In practice evaluate() for NeMo Gym use cases (e.g. OSWorld) inspects the environment state rather than the obs argument, so it is not observable.
4. **`StepError` silently dropped** - added `error: str | None` to `StepResponse`.
5. **Double atexit** - removed `atexit.register`, lifespan is sufficient.
6. **`/tasks` endpoint** - added `GET /tasks` returning `[{idx, task_id}]`.

NicolasAG (approved):
LGTM

PR #92, opened 2026-04-09
feat: uv workspace, parallel tests, integration markers

amanjaiswal73892:
## Summary
- Add `cube-resources/*` as uv workspace members — each sub-package gets `cube-standard = { workspace = true }` replacing individual path deps
- Add `pytest-xdist` to dev deps; set `asyncio_mode = "auto"` and expand `testpaths` to `["tests", "cube-resources"]` for unified discovery from root
- Update `make install`/`make ci-install` to use `--all-packages`; `make test` uses `-n 10`
- Mark `test_local.py` and `test_toolkit.py` with `pytestmark = pytest.mark.integration`, excluded from default run via `addopts = "-m 'not integration'"` — cuts run time from ~5 min to ~22s
- Delete `cube-chat/tests/__init__.py` which caused a module name collision when running pytest from root
- Bump `requires-python` to `>=3.13` in `cube-infra-aws` and `cube-infra-azure`

## Test Plan
- [x] `uv run pytest -n 10` — all tests pass in ~22s
- [x] Integration tests still runnable via `uv run pytest -m integration`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,10 +10,10 @@ help:
 	@echo "make test       - Run unit tests"
 
 install:
-	uv sync --all-extras
+	uv sync --all-extras --all-packages
 
 ci-install:
-	uv sync --frozen --all-extras
+	uv sync --frozen --all-extras --all-packages
 
 update:
 	uv sync --all-extras --upgrade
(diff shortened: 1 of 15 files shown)


amanjaiswal73892:
Landed #91 --> main before #92 --> #72 and the merged conflicts are resolved in #72 --> main.

PR #93, opened 2026-04-09
feat: measure multi-worker throughput in cube test

Shailesh-05:
- Add workers=1 ThreadPoolExecutor path to run_debug_suite; preserve task order
- Document read-only _runtime_context contract after setup for workers>1
- cmd_test: time 1-worker suite, then measure wall-clock runs at 2 and 4 workers
- Replace projected throughput table with Measured/Linear/Efficiency from real runs
- Update stress_test_specs diagram; extend CLI/testing tests

Click the `Preview` tab and select a PR template:

- [Docs/Website-only change](?expand=1&template=docs_website_pr.md)
- [Other change](?expand=1&template=general_code_pr.md)

[//]: # (Dumb that we have to do this hack, see https://github.com/orgs/community/discussions/4620 for progress on github fixing this)


Shailesh-05:
@NicolasAG this LGTM. If you can review and suggest any changes, that would be great.

NicolasAG:
### Code review

Found 3 issues:

1. Multi-worker run results are silently discarded, hiding task failures: `run_debug_suite` is called for `workers=2` and `workers=4` but the return value is never captured. Task failures during parallel runs go unreported, and `n_tasks` (the throughput numerator) is silently reused from the 1-worker run even if the parallel run completes fewer tasks.

https://github.com/The-AI-Alliance/cube-standard/blob/5420c966a8d5ca9c4af2d2823c0aca9175902026/src/cube/cli.py#L605-L612

2. Exceptions from later futures are silently swallowed in the parallel path: iterating `[fut.result() for fut in futures]` in submission order means if `futures[0].result()` raises, Python exits the list comprehension immediately and never calls `.result()` on the remaining futures. Those threads still run to completion (via `ThreadPoolExecutor.__exit__`), but any exceptions they raise are permanently lost. Only the first failure (by submission order) surfaces to the caller.

https://github.com/The-AI-Alliance/cube-standard/blob/5420c966a8d5ca9c4af2d2823c0aca9175902026/src/cube/testing.py#L364-L368

3. Throughput gate uses a different metric than the rate it enables: the guard `if total_ep_s >= _THROUGHPUT_MIN_TOTAL_S` checks the sum of per-episode `episode_time_s` values, but the rate it gates (`rate_1`) is computed from `elapsed_suite_s` — wall-clock time including setup/teardown. These diverge when benchmarks have expensive setup (large `elapsed_suite_s`, small `total_ep_s`) or when tasks do minimal work per episode. Before this PR the rate denominator was also `total_ep_s`, keeping gate and rate consistent.

https://github.com/The-AI-Alliance/cube-standard/blob/5420c966a8d5ca9c4af2d2823c0aca9175902026/src/cube/cli.py#L597-L601

NicolasAG (changes requested):
Claude code spotted a few issues (see above) 


I also ran it on miniwob and the 2 & 4 workers were much much faster than the theoretical linear scale. I'm guessing caching could be the issue here.

-  1-worker rate is measured from `elapsed_suite_s` : the wall-clock time of the very first sequential suite run (cold start, run before the compliance table is even shown).

-  2-worker and 4-worker rates are measured in a second pass, after the first run, so they benefit from warm caches, already-imported modules, and any system-level warm-up.

potential fix:
- Run the 1-worker pass twice, discard the first, use the second as rate_1.

Shailesh-05:
Sure @NicolasAG , noted. I'll work on these changes and let you know shortly.

src/cube/cli.py
```diff
@@ -597,20 +595,47 @@ def cmd_test(
...
+        # Second 1-worker pass (warm): rate_1 uses wall time comparable to 2/4 worker passes below
+        # (the compliance run above is a cold start; discarding it for throughput avoids biased scaling).
+        with console.status(
+            "[info]Warm-up[/info] [dim](1 worker, for throughput baseline)…[/dim]",
+            spinner="dots",
+        ):
+            t_warm0 = time.perf_counter()
+            _ = run_debug_suite(resolved, module, max_steps=max_steps, print_json=False, workers=1)
+            elapsed_warm_1_s = time.perf_counter() - t_warm0
+        rate_1 = n_tasks / (elapsed_warm_1_s / 60.0) if elapsed_warm_1_s > 0 else 0.0
```
NicolasAG: you should probably get the `n_tasks` from the output of `run_debug_suite` to be accurate

NicolasAG (approved):

src/cube/cli.py
Shailesh-05: Sure @NicolasAG, makes sense

PR #97, opened 2026-04-10
feat(chat-tool): add report_infeasible action

recursix:
## Summary

- Add `report_infeasible` action to `ChatTool`, following the same pattern as the existing `send_message` action
- The action posts a message with role `"infeasible"` to the chat session and returns the updated chat history
- This separates communication actions from browser-only actions, aligning with BrowserGym's action model where `send_msg_to_user` and `report_infeasible` are distinct from browser navigation/interaction actions

## Details

The `ChatSession` protocol already supports `role="infeasible"` via `ChatRole = Literal["user", "assistant", "info", "infeasible"]`, so no changes to the session layer were needed. The new action uses `self._session.add_message("infeasible", reason)` rather than `send_message()` since it is a one-way report, not a conversational message that should unblock `wait_for_user_message()`.

## Test plan

- [x] New test: `test_execute_action_dispatches_report_infeasible` -- verifies `add_message` is called with `("infeasible", reason)`
- [x] New test: `test_report_infeasible_returns_observation_with_chat_history` -- verifies the observation contains the formatted chat history
- [x] Updated test: `test_action_set_contains_expected_actions` -- verifies both `send_message` and `report_infeasible` are in the action set
- [x] All pre-existing tests remain unchanged and pass
- [x] `ruff check` and `ruff format` pass with no issues

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cube-tools/cube-chat-tool/src/cube_chat_tool/chat_tool.py
+++ b/cube-tools/cube-chat-tool/src/cube_chat_tool/chat_tool.py
@@ -2,7 +2,7 @@
 
 Provides ``ChatToolConfig`` and ``ChatTool``, a concrete implementation of the
 ``Tool`` base class that wraps a ``ChatSession`` and exposes ``send_message``
-as the single agent-facing action.
+and ``report_infeasible`` as the agent-facing actions.
 """
 
 import datetime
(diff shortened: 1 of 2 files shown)


younik (commented):
report_infeasible is very specific to WorkArena... I wouldn't wire this to ChatTool, but I would make another tool for this? So it is easier to able/disable the action

recursix:
> report_infeasible is very specific to WorkArena... I wouldn't wire this to ChatTool, but I would make another tool for this? So it is easier to able/disable the action

I'm pretty sure WebArena has that as well. I agree that it's a stretch to put in Chat, but it's kind of related. I would be in favor of putting it here now to move fast and avoid having too many different tools.

younik:
> I'm pretty sure WebArena has that as well.

WebArena has submit answer, which includes a param to claim success or failure.



Apologies, I hallucinated that you already agreed on putting it in a separate tool, and I have done it in https://github.com/The-AI-Alliance/cube-harness/pull/279 since it is a WorkArena-specific tool.

If you think this is not a good idea, let's discuss and I can revert, otherwise we can close this!

recursix:
Superseded by The-AI-Alliance/cube-harness#279 (merged), which implements `report_infeasible` as a WorkArena-specific tool rather than wiring it through ChatTool — matching the design direction younik suggested. Closing this one.

PR #99, opened 2026-04-13
Add WAA evaluation methods to GuestAgent

kushasareen:
## Summary

Adds 15 missing methods to `GuestAgent` that the WindowsAgentArena evaluation getters require. The getters call `env.controller.<method>()` where the controller is a `GuestAgent` instance via `GuestAgentProxy`. These methods existed on the original WAA `PythonController` but were never ported.

### Methods added

- `get_vm_documents_path()` — Documents directory path
- `get_vm_folder_exists_in_path()` / `get_vm_file_exists_in_path()` — File/folder existence checks
- `get_vm_are_files_sorted_by_modified_time()` — File sorting check
- `get_file_as_text()` — Download file as text
- `get_vm_file_explorer_is_details_view()` — File Explorer view mode
- `get_file_hidden_status()` — Hidden file attribute check
- `get_vm_is_directory_read_only_for_user()` — Directory permissions
- `get_vm_are_all_images_tagged()` — Image metadata check
- `get_vm_library_folders()` — Windows library folders
- `get_vm_check_if_timer_started()` / `get_vm_check_if_world_clock_exists()` — Clock app checks
- `execute_shell_command()` — Shell command execution via `/execute` endpoint
- `get_registry_key()` — Windows registry via PowerShell
- `get_all_installed_apps()` — Installed apps via PowerShell

Each method follows GuestAgent's existing patterns (retry logic, type hints, timeouts) and calls existing Flask server endpoints.

### Testing

- GuestAgent imports cleanly
- All 24 getter method calls verified present (0 missing)
- Ruff lint passes


cube-tools/cube-computer-tool/src/cube_computer_tool/computer.py
```diff
@@ -61,6 +62,7 @@ class ComputerConfig(ToolConfig):
     cache_dir: str = ""
     require_a11y_tree: bool = True
     require_terminal: bool = False
+    require_obs_winagent: bool = False
     observe_after_action: bool = True
```
amanjaiswal73892, line 66: Document the flags and their effect in the read of the tool.

amanjaiswal73892:
## Code Review Follow-up

I reviewed this PR and applied the following fixes directly to the branch:

### Applied fixes

**EX-001 — Function-scoped imports moved to module level (`azure.py`)**
- Moved `from concurrent.futures import ThreadPoolExecutor, as_completed` out of `cleanup_stale()` and `cleanup_orphaned_resources()` to module top
- Moved `from azure.identity import AzureCliCredential` out of `_CachedCliCredential.__init__` and `model_validator` body to module top

**EX-002 — Documented mutable globals exception**
- Added explanation comments above `_RESERVED_PORTS` in `infra_utils.py` and `_CACHED_CLI_CREDENTIAL` in `azure.py` citing the Ray per-worker isolation rationale (same pattern as the OTel tracer exception in the constitution)

**Security — Replaced `eval()` with `ast.literal_eval()` in `axtree.py`**
Both `screencoord` and `size` parsing now use `ast.literal_eval()` instead of `eval()` on VM-provided data. Added `import ast` at module top.

**`forwarded_ports` mutable default**
Changed `forwarded_ports: list[int] = []` to `forwarded_ports: list[int] = Field(default_factory=list)` in `resource.py`.

**TC-001 — Added `forwarded_ports` to existing RFC deltas**
Added an ADDED entry for `VMResourceConfig.forwarded_ports` to `openspec/changes/resource-convergence/deltas.md`.

---

### Needs discussion (not fixed — too complex for this pass)

**Security — Path injection in `get_file_hidden_status()`**
`file_path` is interpolated directly into a Python command string executed on the remote VM. A path containing a single quote breaks the string. A safe fix requires either a dedicated server-side endpoint or a different parameter-passing mechanism — touches both client and server.

**Security — Plaintext WinRM credential in bootstrap script**
`windows_admin_password` is interpolated into the `_AZURE_BOOTSTRAP_SCRIPT` heredoc and passed as a `curl -u` Basic Auth flag, which exposes it in `ps aux` and `/var/log/cube-bootstrap.log` on the bootstrap VM. The Python-side protection (`exclude=True`, `repr=False`) is correct. Eliminating the log exposure requires an architectural change (Azure Key Vault reference, secrets parameter, or WinRM HTTPS). Suggest filing a follow-up issue.

**CC-002 — No tests for the 15 new `GuestAgent` methods**
No test infrastructure exists for `cube-computer-tool` yet. Worth building — all 15 methods follow the same `requests.post` + retry pattern so a parameterized test suite would cover them efficiently. Suggest a follow-up PR.

amanjaiswal73892 (approved):

NicolasAG (approved):

PR #101, opened 2026-04-14
fix(benchmark): use model_copy instead of deepcopy in subset_from_list

recursix:
## Summary

- `Benchmark.subset_from_list()` was using `copy.deepcopy(self)`, which copies `PrivateAttr` runtime state — subprocess handles, open file descriptors, thread locks — that cannot be pickled.
- The proper fix is to use `self.model_copy(deep=True)`, which copies only Pydantic model fields and leaves all `PrivateAttr`s at their defaults.
- The contract is now explicit in the comment: the returned subset has not been set up; the caller must call `.setup()`.

## Why this is the right fix

`PrivateAttr` runtime state (e.g. `_server_process`, `_stdout_file`) represents live OS resources that belong to the parent benchmark. A subset should never inherit them — it will initialize its own via `setup()`. `model_copy` matches this contract exactly.

Any benchmark with OS-level private attrs (subprocess, file handles, Docker clients, etc.) previously had to override `__deepcopy__` as a workaround. This change removes that need entirely.

## Related

- Removes the need for `MiniWobBenchmark.__deepcopy__` workaround in cube-harness (The-AI-Alliance/cube-harness#282).

## Test plan
- [ ] `benchmark.subset_from_list(task_ids)` works before and after `setup()` on MiniWob
- [ ] Subset contains only the specified tasks
- [ ] Subset can be independently `setup()` and `close()`d


NicolasAG:
### Code review

Both issues are now fixed.

1. ~~**`report_infeasible` does not unblock `wait_for_user_message()` — deadlock**~~ **Fixed in [668617d](https://github.com/The-AI-Alliance/cube-standard/commit/668617d6b438e6ef59e5243ae59e0fe587633869)**

`report_infeasible` called `self._session.add_message("infeasible", reason)`, which appends to message history but does **not** put anything on `BasicChatSession._queue`. Any task thread blocked on `wait_for_user_message()` (which blocks on `_queue.get()`) would hang indefinitely when the agent called this action instead of `send_message`.

Fixed by adding `self._session.stop()` after `add_message`, which puts the stop sentinel on the queue and raises `SessionStoppedError` in any waiting task thread.

https://github.com/The-AI-Alliance/cube-standard/blob/668617d6b438e6ef59e5243ae59e0fe587633869/cube-tools/cube-chat-tool/src/cube_chat_tool/chat_tool.py#L128-L147

---

2. ~~**`model_copy(deep=True)` copies `PrivateAttr` values — the comment and stated rationale are incorrect**~~ **Fixed in [a497c38](https://github.com/The-AI-Alliance/cube-standard/commit/a497c38fb91eca60123b180eae39c9d3b7709e2f)**

`model_copy(deep=True)` copies `__pydantic_private__` along with model fields, so any non-pickleable `PrivateAttr` set during `_setup()` (e.g. `subprocess.Popen` handles and file descriptors in `MiniWobBenchmark`) would have been incorrectly inherited by the subset — or caused a pickling error on copy.

Fixed by explicitly resetting `__pydantic_private__` to fresh defaults after `model_copy`, so the subset starts in a clean unset-up state. Public fields (`resources`, `container_backend`, etc.) are still inherited from the parent. An `info`-level log message is also emitted reminding callers to invoke `.setup()` before use.

https://github.com/The-AI-Alliance/cube-standard/blob/a497c38fb91eca60123b180eae39c9d3b7709e2f/src/cube/benchmark.py#L529-L557

---

NicolasAG (approved):

PR #104, opened 2026-04-15
feat: declarative DockerServiceConfig (VolumeSpec, ServicePort, healthcheck)

recursix:
## Summary

Stacks on #106. Adds the `DockerServiceConfig` and `VolumeSpec` types to the CUBE resource model, implements them in `LocalInfraConfig`, and applies the Generalized VM pattern to the Docker host bootstrap on both AWS and Azure.

**Motivation**: WebArena MAP needs 3 large tarballs downloaded into 9 Docker volumes before the container can start. `launch_script` runs at launch time only — there was no provision-time hook. `VolumeSpec` adds that capability declaratively, so infra backends handle large data downloads once during `provision()` rather than on every `launch()`.

## New types

**`VolumeSpec`** — provision-time Docker volume data:
```python
VolumeSpec(
    name="webarena_map_tile_db",
    mount_path="/data/database",
    source_url="https://...osm_tile_server.tar",
    tar_subpath="projects/.../osm-data/_data",
    strip_components=6,
)
```
Fields: `name`, `mount_path`, `source_url?`, `tar_subpath?`, `strip_components` (default 0).

**`DockerServiceConfig`** — multi-image, multi-container, multi-port resource:
```python
DockerServiceConfig(
    name="webarena-shopping-admin",
    scope="benchmark",
    docker_images=["am1n3e/webarena-verified-shopping_admin"],
    services={"shopping_admin": 7780, "shopping_admin_ctrl": 7781},
    volumes=[...],                          # VolumeSpec list, empty = no volume data
    endpoint_to_site={"shopping_admin": "shopping_admin"},
    launch_script="...",                    # bash run on the Docker host at launch
)
```

## LocalInfraConfig changes

`LocalInfraConfig` now supports `DockerServiceConfig` through the same `provision()` / `launch()` interface:

- **`provision()`**: `docker pull` each image + runs `build_volume_setup_script()` for any `VolumeSpec` entries.
- **`launch()`**: runs `launch_script` via bash locally; snapshots `docker ps -q` before/after to identify new containers; builds `endpoints` as `{service_name: "http://127.0.0.1:{port}"}`.
- **`LocalDockerServiceHandle.close()`**: `docker stop` + `docker rm` every tracked container + removes the `active.json` entry.
- **Bug fix**: `ProvisionStore().set()` → `.put()` (`.set()` does not exist — provision always raised `AttributeError`).

## Generalized VM pattern (AWS + Azure)

Extends the fix from #105 to the Docker host bootstrap scripts:

- SSH public key no longer baked into the provisioned AMI / Gallery image.
- AWS: `_AWS_DOCKER_BOOTSTRAP_SCRIPT` no longer has a `{ssh_pubkey}` placeholder.
- Azure: `_DOCKER_BOOTSTRAP_SCRIPT` runs `waagent -force -deprovision+user` before the sentinel, producing a Generalized image. Username changed `azureuser` → `cube` to match the VM pattern.

## Infra utility

`build_volume_setup_script(volumes)` in `cube/infra_utils.py` generates an idempotent bash script that downloads + extracts archives into Docker volumes. Shared by LocalInfraConfig and the cloud backends.

## Test plan

- [x] 191/191 cube-standard unit tests pass
- [x] Azure integration test: full provision → launch → WAV debug suite → unprovision (exit 0) ✅
- [x] AWS integration test: AMI provisioned successfully with Generalized VM pattern ✅
- [x] LocalInfraConfig integration test: provision → launch → HTTP 200 → close → container gone ✅ (cube-harness#295)


NicolasAG:
### Code review

Found 1 issue:

1. `_kill_entry()` and `list_active()` are not updated to handle `docker_service` entries — Docker containers leak on process crash

`_launch_docker_service()` writes `{"type": "docker_service", "container_ids": [...], ...}` to `active.json`, but the two cleanup paths that handle crashes bypass `LocalDockerServiceHandle.close()` entirely:

- `_kill_entry()` only handles `pid` (SIGTERM/SIGKILL) and `overlay_path` removal — it never reads `container_ids` and never calls `docker stop`/`docker rm`. Any docker service entry passed to it silently leaks its containers.
- `list_active()` reconstructs all entries unconditionally as `VMResourceConfig` + `LocalResourceHandle`, regardless of the `type` field. Docker service entries get a handle whose `close()` does nothing useful (no pid, no overlay, no containers).

Both `cleanup()` (line 585) and `cleanup_stale()` (line 620) call `_kill_entry()` — these are the exact paths intended to catch containers after a process crash or TTL expiry. They will deregister the `active.json` entry but leave the Docker containers running indefinitely.

https://github.com/The-AI-Alliance/cube-standard/blob/f1979179c40a3d5bc8bde8c9407420ed9206eca0/src/cube/infra_local.py#L643-L662

https://github.com/The-AI-Alliance/cube-standard/blob/f1979179c40a3d5bc8bde8c9407420ed9206eca0/src/cube/infra_local.py#L548-L565

recursix:
Good catch — both bugs confirmed and fixed in 3f0a063.

**`_kill_entry()`**: now reads `container_ids` from the entry and runs `docker stop` + `docker rm` on each before the existing pid/overlay logic.

**`list_active()`**: branches on `entry.get("type") == "docker_service"` and reconstructs a `LocalDockerServiceHandle` (with `container_ids` and `endpoints`) instead of blindly creating a `LocalResourceHandle` for every entry.

`cleanup()` and `cleanup_stale()` both call `_kill_entry()` so they now correctly stop Docker containers on TTL expiry or explicit cleanup.

NicolasAG (approved):

PR #107, opened 2026-04-16
Add WebSearchTool & WebFetchTool

younik:
Adds WebSearchTool & WebFetchTool for DeepResearch benchmarks (e.g. BrowserComp)


cube-tools/cube-web-tool/src/cube_web_tool/fetch_tool.py
```diff
@@ -0,0 +1,75 @@
...
+            logger.warning("web_fetch markitdown failed for %s: %s", url, fetch_err)
+            try:
+                response = self._http_client.get(url)
+                response.raise_for_status()
+                markdown = response.text
+            except Exception as e:
+                return f"Fetch error: {e}"
+
+        truncated = markdown[: self.config.max_fetch_chars]
+
```
kushasareen, line 68: I'm wondering why the fetch LLM model should be part of the tool and potentially different from the model we are evaluating? Wouldn't it make more sense to provide the web markdown directly to the agent and let it extract the relevant info? 

It can make sense for context management reasons but that can also be handled by agent design. I'm not sure why this is a good choice or if it's standard for these types of benchmarks.

cube-tools/cube-web-tool/tests/test_web_tool.py
```diff
@@ -0,0 +1,43 @@
+"""Tests for WebSearchTool and WebFetchTool."""
+
+from cube_web_tool import BraveWebSearchToolConfig, WebFetchToolConfig
+
```
kushasareen, line 9: Maybe we can have a test that checks if search result are actually returned?

kushasareen (commented):
Other than these points it looks good to me!

cube-tools/cube-web-tool/src/cube_web_tool/search_tool.py
```diff
@@ -0,0 +1,87 @@
...
+        self._http_client.close()
+
+    def _resolve_brave_api_key(self) -> str:
+        if self.config.brave_api_key is not None:
+            return self.config.brave_api_key
+        key = os.environ.get("BRAVE_API_KEY")
+        if key is None:
+            raise ValueError("BRAVE_API_KEY not set in config or environment")
+        return key
+
```
kushasareen, line 60: I think we should need to declare @tool_action here right?

kushasareen (changes requested):

cube-tools/cube-web-tool/src/cube_web_tool/search_tool.py
```diff
@@ -0,0 +1,87 @@
...
+    """Tool providing Brave Search for cube benchmarks."""
+
+    def __init__(self, config: BraveWebSearchToolConfig) -> None:
+        self.config = config
+        self._http_client = httpx.Client(timeout=30.0, headers={"User-Agent": _USER_AGENT})
+
+    def close(self) -> None:
+        self._http_client.close()
+
+    def _resolve_brave_api_key(self) -> str:
```
kushasareen, line 53: This should also consider when BRAVE_API_KEY is "", which was the case when I ran it and and by default. This currently slips through

cube-tools/cube-web-tool/src/cube_web_tool/search_tool.py
kushasareen: It also needs a litellm docstring I guess

cube-tools/cube-web-tool/src/cube_web_tool/fetch_tool.py
younik: Yes, it is mainly for efficient context management.
If we handle it in the agent, we should ensure this doesn't bloat our context, which we don't have atm. 
We can achieve this by allowing the agent to spawn a subagent to extract relevant information from a tool, but will add a bit of more workload to the main agent that must decide when this should be used, and will affect all other tools.

For reference, the current implementation reflects the interface of Claude Code; I am fine in doing it otherwise, but it requires a complete eval on the benchmarks.

cube-tools/cube-web-tool/tests/test_web_tool.py
younik: Calling the search tool requires a Brave API key, so I can only create a mock of API calls

cube-tools/cube-web-tool/src/cube_web_tool/search_tool.py
younik: as it extends WebSearchTool, that already have `@tool_action`, there is no need for that

younik: same for docstring, they are inherited from the base class

cube-tools/cube-web-tool/tests/test_web_tool.py
kushasareen: Ok, this is fine.

cube-tools/cube-web-tool/src/cube_web_tool/fetch_tool.py
kushasareen: I agree and think the current way would probably get higher results. I guess there's a couple broader questions here:

1. Should we try to exactly implement benchmarks or improve them?
2. When a model better than gpt-5.4 comes out for summarization and we eval that model, would this suppress the results?
3. I guess an agent should probably be able to get access to the web fetch and see everything if it wants to but that isn't allowed currently. There's always an option to add another tool for that and let the user choose, which might be the best thing.

I'd feel generally more confident if could see how the results were.
Maybe @NicolasAG has some input?

cube-tools/cube-web-tool/src/cube_web_tool/search_tool.py
kushasareen: image

That doesn't seem true. Maybe I'm missing something. In any case, it should be made explicit.

kushasareen: Ok I see where the tool_action logic happens at runtime in action_set. But I don't think the docstring is inherited.

The original design choice is strange to me since it should be explicit what is/isn't a tool action without looking all the way up the inheritance tree. Also, if that's the norm then I guess the docstring should be inherited too but I don't see that anywhere.

I think the docstring still needs to be fixed.

cube-tools/cube-web-tool/src/cube_web_tool/fetch_tool.py
NicolasAG: This is an interesting discussion. I think it is valuable to offer the agent the flexibility of doing both: return raw truncated md or return litellm match.
So we can either add a bool flag to the web_fetch tool action, something like this:
```python
@tool_action
def web_fetch(self, url: str, query: str, match_query: bool = True) -> str:

    # ... fetch markdown ...

    if not match_query:
        return markdown[: self.config.max_fetch_chars]
    else:
        # ...
```
or have two separate actions:
```python
@tool_action
def url_to_md(self, url: str) -> str:
    # ... fetch markdown ...
    return markdown[: self.config.max_fetch_chars]

@tool_action
def web_fetch(self, url: str, query: str) -> str:
    markdown = self.url_to_md(url)
    # ...
```

younik: Maybe we can let the Query be optional to achieve the same result, or we can have two different tools? So we let the user decide how this should be handled?

cube-tools/cube-web-tool/src/cube_web_tool/search_tool.py
younik: The doc is inherited:
```
uv run --with ./cube-tools/cube-web-tool python -c "import inspect; from cube_web_tool.search_tool import BraveWebSearchTool; print(inspect.getdoc(BraveWebSearchTool.web_search))"
```
gives: 
```
Search the web and return a list of results.

Args:
    query: The search query string.
```

I get your point on readability tho; I don't mind repeating them.

younik: Done it https://github.com/The-AI-Alliance/cube-standard/pull/107/commits/1f4cf82d6ed4b8bdaa5d8c0eb6965585a5f1ab0b

cube-tools/cube-web-tool/src/cube_web_tool/search_tool.py
younik: > and by default.

I don't get this, if you unset BRAVE_API_KEY, you should get None, isn't it?

cube-tools/cube-web-tool/src/cube_web_tool/fetch_tool.py
kushasareen: I agree with that

cube-tools/cube-web-tool/src/cube_web_tool/search_tool.py
kushasareen: Thanks for clarifying and the changes. I learned something new. Looks great :)

cube-tools/cube-web-tool/src/cube_web_tool/search_tool.py
kushasareen: Yea, that's my bad. I had it set as "" in .env. LGTM

cube-tools/cube-web-tool/src/cube_web_tool/fetch_tool.py
younik: I have added support for fetching without a query, based on WebFetchToolConfig.query_llm_model, if it is set or not https://github.com/The-AI-Alliance/cube-standard/pull/107/commits/7b3f13d896e55f44dd9560732a2adaf18630f778

kushasareen: Happy to approve now

kushasareen (approved):
Good

PR #118, opened 2026-04-22
feat(benchmark): BenchmarkConfig split + CompositeBenchmark (closes #96 #111 #122)

### LINKED ISSUE #96: Support runtime InfraConfig injection in cube testing/debug flows
## Summary

The current `cube.testing` debug flow assumes a zero-argument `get_debug_benchmark()` and does not provide a clean way to inject a runtime-constructed `InfraConfig`.

This makes it awkward to run debug tasks programmatically or from CLI when the benchmark needs a non-default infra instance created by the caller.

## Problem

Today, helpers like `run_debug_suite(...)` call:

```python
benchmark = module.get_debug_benchmark()
```

That means:
- the caller cannot pass a prebuilt `InfraConfig`
- benchmarks must rely on defaults, env vars, or benchmark-local logic
- adding support for new infra types pushes orchestration concerns into benchmark packages

## Goal

Add a framework-supported way to provide `InfraConfig` at runtime for debug/testing flows without requiring benchmark-specific env-var handling.

## Possible directions

- Add a `cube.testing` helper that accepts a benchmark instance directly
- Add support for a benchmark factory callable instead of a zero-arg module convention
- Add CLI support for runtime infra injection, such as an infra factory/import path
- Pass resolved infra through runtime context where appropriate

## Why this matters

This keeps benchmark packages infra-agnostic and moves orchestration into the framework layer, where it belongs.

It is especially useful for benchmarks like OSWorld where tasks require a caller-provided infra backend.

## Acceptance criteria

- A caller can run debug/testing flows with a runtime-constructed `InfraConfig`
- Benchmarks do not need provider-specific logic in `debug.py`
- The solution works both programmatically and is compatible with future CLI improvements

## Related

- #111 — `BenchmarkConfig.make(infra)` introduced there is the natural factory mechanism that solves runtime infra injection; a single PR likely resolves both

### LINKED ISSUE #111: feat: BenchmarkConfig and make() API
## Context

RFC [cube-harness#294](https://github.com/The-AI-Alliance/cube-harness/pull/294) proposes splitting `Benchmark` into a serializable config layer and a runtime layer, mirroring the existing `TaskConfig` / `Task` pattern:

```
BenchmarkConfig  →  make(infra)  →  Benchmark
  serializable                      not serializable
  subset_from_list()                _setup() / close()
```

This unlocks clean composition (`CompositeBenchmark`) and parallelism (`BenchmarkPool`) in the harness — but the `BenchmarkConfig` / `make()` contract must be defined and landed in cube-standard first.

## What needs to happen here

- [ ] Define `BenchmarkConfig` as a pure Pydantic model in cube-standard (alongside `Benchmark`)
- [ ] Add `BenchmarkConfig.make(infra) -> Benchmark` factory method
- [ ] Move `subset_from_list` / `subset_from_glob` onto `BenchmarkConfig` (pure data copy, no deepcopy hacks)
- [ ] Update / add tests for the round-trip: `BenchmarkConfig` → serialize → deserialize → `make()` → `Benchmark`
- [ ] Decide and document the cube-standard vs harness boundary (what lives here vs in the harness)

## Why

- Eliminates the `ClassVar` / deepcopy workarounds currently in `Benchmark.subset_from_list`
- Makes the full benchmark suite serializable so the harness can dispatch it across Ray workers
- Required prerequisite for `BenchmarkPool` and `CompositeBenchmark` in the harness

## Related

- RFC: The-AI-Alliance/cube-harness#294
- #96 — `BenchmarkConfig.make(infra)` is the runtime infra injection mechanism that issue asks for; a single PR likely resolves both

### LINKED ISSUE #122: Conflict between Benchmark.__init_subclass__() and Benchmark.install()
Suppose `MyBenchmark` is a subclass of `Benchmark`, and `MyBenchmark.install()` is expected to create `task_metadata.json`. However, `Benchmark.__init_subclass__()` is called before `MyBenchmark.install()`, and it requires `task_metadata.json` to already exist.

This means there is a conflict:

- If `MyBenchmark.task_metadata` is not defined, and `task_metadata.json` does not exist at the beginning, `Benchmark.__init_subclass__()` will raise error.

- If `task_metadata.json` already exists at the beginning, it will be covered by the new file created by `MyBenchmark.install()`.

NicolasAG:
## Summary

Implements the `BenchmarkConfig` / `Benchmark` split described in [cube-harness/docs/rfc-benchmark-config-and-scaling.md](https://github.com/The-AI-Alliance/cube-harness/blob/main/docs/rfc-benchmark-config-and-scaling.md) and adds `CompositeBenchmarkConfig` for cross-benchmark composition. This is the main unlock behind #111 and resolves the infra-injection gap in #96.

**504 tests pass, 5 skipped. No regressions.**

## What's in this PR

| Commit | Scope |
|---|---|
| `c11ce38` | openspec change folder (`proposal.md` + `deltas.md`) |
| `c7e3536` | Core `Benchmark` → `BenchmarkConfig` (Pydantic) + `Benchmark` (plain class) split; `get_debug_benchmark()` protocol; counter-cube migrated; all specs updated |
| `b5640b8` | `CompositeBenchmarkConfig` + `CompositeBenchmark`; 16 dedicated tests |
| `18bd34a` | **Option 1** — embed `TaskMetadata` + `ToolConfig` directly in `TaskConfig`; `sub_bench_name` replaces `CompositeTaskConfig` wrapper |
| `6dd1985` | Rename `sub_benchmark_name` → `sub_bench_name`; `sub_configs` → `sub_bench_configs` |
| `508b514` | `make_benchmark_rpc_server` uses a real subprocess via `BenchmarkConfig`; both entry points take JSON strings |
| `078e710` | JSON round-trip enforced at every cross-process boundary: `SerializeAsAny` on all polymorphic fields; `AbstractSeedGenerator → TypedBaseModel`; `_dump_/_load_runtime_context` helpers for `InfraConfig`-containing dicts |
| `e422581` | `cube.benchmarks` entry-point group now advertises `BenchmarkConfig` classes; `get_debug_benchmark()` takes no args (infra belongs at `config.make(infra)`, not at config construction) |
| `e6f7dd1` | `_template` scaffold updated to the `BenchmarkConfig`/`Benchmark` pair shape |
| `2e8711e` | Archive `openspec/changes/benchmark-config/` → `openspec/changes/archive/2026-04-24-benchmark-config/`; apply remaining deltas to main specs |
| `d71217d` | `.claude/skills/new-cube` and `.claude/skills/review-cube` updated to reflect all new design decisions |

## Key design decisions

1. **Two classes**, mirroring `TaskConfig`/`Task`: `BenchmarkConfig` holds serializable data + class-level registries; `Benchmark` is a plain Python class holding only live OS state.
2. **`task_metadata` stays a ClassVar** (authoritative registry); subsets narrow via `task_ids: list[str] | None` on the config instance — no more `__pydantic_private__` hacks.
3. **Infra at `make(infra)` time, not config construction** — `get_debug_benchmark()` returns a pure config, no args needed. Cubes that need a persistent infra field (e.g. swebench-live-cube's `infra: InfraConfig`) declare it as a typed Pydantic field on their `BenchmarkConfig` subclass; `_setup()` reads `self.config.infra`.
4. **Option 1 for TaskConfig** — `TaskConfig` carries its own `metadata: SerializeAsAny[TaskMetadata]`, stamped by `get_task_configs()` on the driver. Workers are self-contained — no import of the owning `BenchmarkConfig` needed. `CompositeTaskConfig` wrapper eliminated; composite routing uses `sub_bench_name: str | None` on the base `TaskConfig`.
5. **`make()` is eager** — provisions resources, calls `setup()`, returns a ready `Benchmark`. Context-manager form (`with config.make() as bench`) handles `close()` automatically.
6. **JSON round-trip at every process boundary** — `SerializeAsAny` on all polymorphic fields; `_dump_/_load_runtime_context` serializes `TypedBaseModel` values (e.g. `InfraConfig` subclasses) inside `runtime_context` dicts. Subprocess entry points receive JSON strings, never live Pydantic objects.
7. **Composites nest freely** — `sub_bench_configs: list[SerializeAsAny[BenchmarkConfig]]` preserves subclass fields through JSON round-trip.

## Open design questions (follow-up to be adressed in #121)

- **`extra_info` split**: currently `TaskMetadata.extra_info` is a single dict accessible to everyone. Cubes like swebench and terminalbench use it for heavy evaluation data (problem statements, patches, test commands) that is not needed at task-inspection time. Should this be split into two distinct fields — one for lightweight public metadata (always loaded) and one for heavy execution data (lazy / loaded only at task construction time)? The swebench pattern of merging `load_task_execution_info()` into `extra_info` at `get_task_configs()` emission time is the current workaround. Worth a dedicated design discussion before more cubes adopt the same pattern.

## Migration notes

- **`AbstractSeedGenerator` is now a `TypedBaseModel`** (changed from a plain `BaseModel` in `078e710`). This is a **silent breaking change**: cubes with a custom seed generator subclass will deserialize on workers without the `_type` discriminator and fail with a `_type`-not-found error. Cube authors must audit their seed generators during migration and ensure subclasses survive a JSON round-trip via `TypedBaseModel.model_validate(...)`.

## Out of scope

- `BenchmarkPool` (deferred to a follow-up PR).
- Migration of cube-harness cubes (OSWorld, WorkArena, SWE-bench, etc.) — that work happens in parallel in separate PRs on cube-harness.

## Test plan

- [x] `make test` passes (504 tests, 5 skipped)
- [x] `make lint` clean
- [x] Composite tests: duplicate-name guard, prefixed `task_metadata`, `sub_bench_name` routing, spawn routing, close resilience, JSON round-trip including composite-of-composite, `install()` delegation
- [x] Server tests: subprocess isolation, `_dump_/_load_runtime_context` round-trip, non-JSON-serializable value rejection
- [x] openspec change archived; spec deltas applied to `testing/spec.md` and `cli/spec.md`
- [x] `.claude/skills/new-cube` and `.claude/skills/review-cube` updated
- [ ] Per-cube smoke tests on cube-harness (deferred to parallel PRs)

Closes #96, closes #111.

Related: cube-harness RFC


NicolasAG:
## Design discussion: where does `task_metadata` live?

Recording this because the first three commits on this branch took one path, and after further reflection I think we should take a different one. Leaving a trail so reviewers understand the shift.

### The question

`TaskConfig.make()` runs on a worker and needs `TaskMetadata` for its `task_id` to construct the Task. Where does that metadata come from?

### Starting point (pre-PR)

```python
# Pre-PR — the existing Benchmark class held everything as ClassVars:
class OSWorldBenchmark(Benchmark):
    task_metadata: ClassVar[dict[str, TaskMetadata]] = {...}
    ...

class OSWorldTaskConfig(TaskConfig):
    def make(self, runtime_context=None, container_backend=None):
        from osworld_cube.benchmark import OSWorldBenchmark  # import inside function to avoid circular imports
        metadata = OSWorldBenchmark.task_metadata[self.task_id]  # ClassVar lookup
        ...
```

`subset_from_list` on the old `Benchmark` tried to create a new instance with a narrowed `task_metadata`. Because `task_metadata` was a `ClassVar`, Pydantic v2 refused to let it be overridden on an instance — the existing code worked around this with `copy.deepcopy(ClassVar) + object.__setattr__(new_instance, 'task_metadata', ...)`.

### Options considered

#### Option 1 — Embed `TaskMetadata` directly in each `TaskConfig`

```python
class TaskConfig(TypedBaseModel, ABC):
    task_id: str
    metadata: TaskMetadata           # travels with the config
    seed: int | None = None
    tool_config: ToolConfig | None = None
```

TaskConfig is fully self-contained. Workers don't import the owning BenchmarkConfig class. No circular dependencies. `BenchmarkConfig` still has a `task_metadata` registry, but it's only used at `get_task_configs()` emit time to stamp metadata onto each emitted config. Subsetting tracked via `task_ids: list[str] | None` on the BenchmarkConfig instance.

#### Option 2 — ClassVar registry + `task_ids: list[str] | None`

```python
class BenchmarkConfig(TypedBaseModel, ABC):
    task_metadata: ClassVar[dict[str, TaskMetadata]]
    task_ids: list[str] | None = None     # subset marker
```

Workers still do `OSWorldBenchmarkConfig.task_metadata[self.task_id]` — class-level lookup is stable across subsets because subsetting only narrows `task_ids`, never touches the ClassVar. Cleaner than the `object.__setattr__` hack, doesn't bloat TaskConfig.

#### Option 3 — Pass `BenchmarkConfig` into `make()`

`TaskConfig.make(config, runtime_context, container_backend)`. Heavy signature; worker deserializes the whole owning config just to look up one task's metadata. Ruled out early.

#### Option 4 — Module-level variable

```python
# my_cube/__init__.py
TASK_METADATA: dict[str, TaskMetadata] = load_from(__file__)

# my_cube/task.py
from my_cube import TASK_METADATA
class MyTaskConfig(TaskConfig):
    def make(self, ...):
        metadata = TASK_METADATA[self.task_id]
```

Decouples the registry from the class. Eliminates the ClassVar-on-subclass tension entirely. Costs: loses `__init_subclass__` validation (forgotten `TASK_METADATA` only fails at first worker call), fragments per-cube surface (one class + a module constant instead of just one class).

### What we did in commits 1–3

**Option 2.** Reasoning at the time: "smaller refactor, smaller blast radius." Tests pass, 464 green, composite works. The ClassVar/instance tension surfaces only in `CompositeBenchmarkConfig`, which we patched with:

- `_skip_init_subclass_checks: ClassVar[bool] = True`
- `@property` overrides of `task_metadata` / `benchmark_metadata` to shadow the parent's ClassVar annotation
- `SerializeAsAny[BenchmarkConfig]` on `sub_configs` to preserve subclass-specific fields through nested JSON round-trips
- A `CompositeTaskConfig(sub_name, inner)` wrapper whose `make()` delegates to `inner.make()`

Those ~40 lines of descriptor gymnastics are the single ugliest part of the PR. They exist entirely because TaskConfig depends on a class-level lookup.

### Why we're switching to Option 1

Revisiting with fresh eyes, Option 2's \"smaller refactor\" argument was a scope optimization, not a design optimization. Given a blank slate:

1. **TaskConfig becomes fully self-describing.** No import of the owning BenchmarkConfig class on workers. No deferred-import dance to avoid circulars. The worker needs only the TaskConfig subclass for `_type` polymorphic dispatch.

2. **`install()` / `load_task_execution_info` merge happens once on the driver**, inside `BenchmarkConfig.get_task_configs()`. Today task config workers opens a per-task JSON file — that disappears. The TaskConfig arrives complete.

3. **Composite gets radically simpler.** `CompositeTaskConfig` can probably be deleted. Composite's `get_task_configs()` calls each sub's `get_task_configs()` and clones each emitted TaskConfig with `task_id=f\"{sub}/{tc.task_id}\"`. `CompositeBenchmark.spawn(tc)` just calls `tc.make(runtime_context=sub_benchmarks[...]._runtime_context, …)`. **Gone:** the wrapper class, `_skip_init_subclass_checks`, `@property` shadowing, `SerializeAsAny` gymnastics, `inner.make()` delegation.

4. **The cost I originally worried about is negligible.** ~200–500 bytes of TaskMetadata per task. Never a bottleneck.

5. **Principled.** TaskConfig's job is to be the serializable handle that becomes a Task on a worker. Metadata is part of the Task definition. It should be on TaskConfig.

We'll land the switch in a subsequent commit on this branch before the remaining work (server subprocess, CLI entry-points, template, archive). The change is mechanical: add `metadata: TaskMetadata` to TaskConfig, update each cube's `.make()` to use `self.metadata`, delete `CompositeTaskConfig`, simplify the composite tests.

### What a 100x engineer would do *(to be revisited when we actually scale up experiments)*

Option 1 plus a `benchmark_ref` field carrying `{package, version}`:

```python
class TaskConfig(TypedBaseModel, ABC):
    task_id: str
    metadata: TaskMetadata
    benchmark_ref: BenchmarkRef      # {package: \"osworld-cube\", version: \"1.0.0\"}
    seed: int | None = None
    tool_config: ToolConfig | None = None
```

Why this pays off at 100x scale (not now, but in the design shape):

1. **Composite routing becomes routing-by-provenance.** No `sub_name` field, no wrapper. `CompositeBenchmark.spawn(tc)` dispatches on `tc.benchmark_ref` against a `dict[BenchmarkRef, Benchmark]`.

2. **Trajectory records self-describe.** At scale the thing that grows fastest isn't config bytes — it's stored trajectories, evaluation artifacts, cross-experiment diffs. Every stored record references a TaskConfig; if the config carries `benchmark_ref`, every row in every log for the next 5 years is self-describing without joins against a mutable registry.

3. **Version-gating falls out of the data.** A worker receiving `benchmark_ref.version=\"1.2.0\"` while it has `osworld-cube==1.3.0` installed can refuse, warn, or upgrade — from the config alone, no out-of-band coordination.

4. **It's the hook point for a registry if embedding ever becomes expensive.** Swap `metadata: TaskMetadata` → `metadata: TaskMetadata | None` + `resolve(registry)` method that fills from `benchmark_ref`. API boundary stays stable.

**Deliberately deferred.** We will not build a registry, a content-addressed store, or a lookup protocol now — those are premature. The `benchmark_ref` field alone is enough to keep the door open. Revisit when (a) trajectory storage becomes a real concern, (b) cross-version coordination becomes a real concern, or (c) config serialization size becomes a measurable bottleneck. File an issue against this PR once merged so the decision is explicit next time.

NicolasAG:
## Correction to the earlier design discussion

I overstated what would disappear from the composite implementation after switching to Option 1. Posting the actual accounting now that commit 18bd34a has landed the refactor.

### What I claimed would be gone

From my earlier comment:

> **Gone:** the wrapper class, `_skip_init_subclass_checks`, `@property` shadowing, `SerializeAsAny` gymnastics, `inner.make()` delegation. All of it.

### What's actually gone

- ✅ **`CompositeTaskConfig` wrapper class** — deleted. Composite now emits each sub-config's native TaskConfig subclass unchanged in type, with only `task_id` (prefixed) and `sub_benchmark` (tag) updated via `model_copy(update=...)`.
- ✅ **`inner.make()` delegation** — gone, no wrapper to delegate through.
- ✅ **`_skip_init_subclass_checks` opt-out flag** — removed. `__init_subclass__` is now descriptor-aware: if a subclass declares `benchmark_metadata` / `task_metadata` as a `@property`, validation steps out of the way. `task_config_class` is only required when the subclass uses the default `get_task_configs`.

### What's still there (and why)

- ⚠️ **`@property` on composite's `benchmark_metadata` / `task_metadata`** — still present. This is the unavoidable consequence of wanting composites to be `BenchmarkConfig` subclasses (so they compose recursively via `sub_configs: list[BenchmarkConfig]`). Standard cubes use `ClassVar` for static registries; composites need dynamic per-instance computation. Same public API (`config.benchmark_metadata`), two legitimate implementations.

  The only way to remove the shadow would be an abstract-base split — introduce an internal `_BenchmarkConfigBase(ABC)` with `benchmark_metadata` as an abstract `@property`, then have `BenchmarkConfig` (standard cubes, ClassVars) and `CompositeBenchmarkConfig` (composites, @property) as sibling concrete implementations. It adds a base class users never interact with directly and leaks into `sub_configs: list[_BenchmarkConfigBase]` hints. Costs more than it buys.

- ⚠️ **`SerializeAsAny[BenchmarkConfig]` on composite `sub_configs`** — still present. My earlier \"gone\" claim was simply wrong; this has nothing to do with where TaskMetadata lives. Pydantic v2 requires `SerializeAsAny[T]` on any polymorphic list field to preserve subclass-specific fields through dump/validate round-trips. Since `sub_configs` genuinely is polymorphic (`OSWorldBenchmarkConfig` and `WorkArenaBenchmarkConfig` have different Pydantic fields), we need it.

### Takeaway

The Option 1 refactor delivered most of the simplification I promised — the wrapper class, delegation, and opt-out flag are all gone — but not every line of ceremony. I think the remaining `@property` + `SerializeAsAny` are the right trade at this scope; deeper removal would be a bigger refactor than it's worth right now. Flagging this transparently so the decision is on the record.

NicolasAG:
Pushed 6 commits addressing the code-review findings:

- [`ef48723`](https://github.com/The-AI-Alliance/cube-standard/commit/ef487231a92ca72a4a93f0ed7e58d10e85e69cef) `fix(testing)`: read `container_backend` from `config` (not `benchmark`) in `check_reset_reproducibility` — the previous `getattr(benchmark, ...)` silently returned `None`.
- [`ba0af2f`](https://github.com/The-AI-Alliance/cube-standard/commit/ba0af2fc6f3dde5cf3d05126fe35ac180bf0e1bf) `fix(nemogym)`: call `config.install()` instead of `type(config).install()` so `CompositeBenchmarkConfig`'s instance-method override dispatches correctly.
- [`64a589b`](https://github.com/The-AI-Alliance/cube-standard/commit/64a589b6ea22b7eda3a5e6ddec02bcfba3893d0d) `docs(spec)`: record the `container_backend` deprecation in the canonical benchmark spec and the archived BenchmarkConfig delta, with the migration path to `resources`.
- [`09b4f9a`](https://github.com/The-AI-Alliance/cube-standard/commit/09b4f9af9e7fafa6f3de8e7bca57b4c2c1e2f2fa) `fix(benchmark)`: restore the post-filter empty-result guard in `subset_from_list` that the refactor dropped.
- [`1b9153d`](https://github.com/The-AI-Alliance/cube-standard/commit/1b9153d3dd0cb3ddb50d24e29e85b6a2cda23d33) `fix(examples)`: wrap `test_counter_benchmark` in `with config.make() as benchmark:` and track tasks in a `finally` so an early assertion failure no longer leaks the benchmark or any task.
- [`6641aa9`](https://github.com/The-AI-Alliance/cube-standard/commit/6641aa911b97e8c5c9a884b32d33a8c9d0050a45) `docs`: update `load_task_execution_info` docstring (driver-side, not workers) and correct the archived `get_debug_benchmark` signature to the zero-arg form that actually shipped.

`make test` still green: 505 passed, 5 skipped.

amanjaiswal73892:
**Code Review**

The design is solid — the `BenchmarkConfig`/`Benchmark` split mirrors the existing `TaskConfig`/`Task` pattern cleanly, `SerializeAsAny` is applied consistently across all polymorphic fields, and the subprocess+JSON boundary in `make_benchmark_rpc_server` is a meaningful correctness improvement over the old thread approach. Composite routing via prefixed task IDs is elegant and the test coverage (especially the round-trip and composite-of-composite tests) is thorough.

The 6 commits pushed today address several of the issues I found in the initial pass:

- `ef48723` (`testing.py` container_backend fix), `ba0af2fb` (nemogym `config.install()`) — both were real bugs that would have been silent failures in practice.
- `64a589b1` (container_backend deprecation docs) and `6641aa98` (docstring/delta alignment) — fill the documentation gaps.
- `09b4f9ac` (empty-result guard in `subset_from_list`) and `1b9153d0` (context manager in counter example) — correct hygiene fixes.

**Still open (before merge)**

**1. Metadata stamping has no enforcement** — `BenchmarkConfig.get_task_configs()` stamps `metadata` on each config, but a subclass that overrides the method and forgets the stamp will produce a worker crash with a confusing "field required" error, not a helpful "you forgot to stamp metadata" message. No mechanism currently prevents this — not a validator, not a post-yield assertion, not an `__init_subclass__` check. At minimum, a docstring warning on the method would help; a runtime guard in the base implementation would be better.

**2. `AbstractSeedGenerator` → `TypedBaseModel` is a silent breaking change** — Any cube with a custom seed generator subclassing the old `BaseModel`-based `AbstractSeedGenerator` will fail at deserialization with a cryptic `_type`-not-found error. This isn't called out in the PR description or migration notes. Worth adding a callout so cube authors know to audit their seed generators during migration.

**Lower priority (follow-up acceptable)**

- `CompositeBenchmark.spawn()` calls `task_config.make()` directly rather than `sub_bench.spawn()` to avoid the prefixed-id rejection — intentional and correct, but a one-line comment explaining *why* would save future readers a confusing detour.
- `_load_runtime_context._rehydrate()` recurses into nested dicts/lists with no depth limit. In practice, runtime contexts are shallow and acyclic, but a comment noting that assumption would make the invariant explicit.
- Tests for seed-generator expansion through the full BenchmarkConfig → JSON → worker pipeline, and for `_dump_runtime_context` raising `TypeError` on non-JSON-native values, are still absent.

NicolasAG:
Pushed 5 more commits addressing the second-pass review:

- [`fc1a8ed`](https://github.com/The-AI-Alliance/cube-standard/commit/fc1a8edd45e0a17fbd3fb87f6acc3b5c54dc52b1) `fix(task)`: add a `model_validator(mode="after")` on `TaskConfig` that asserts `metadata.id` matches `task_id` (prefixed for composites) — a custom `get_task_configs()` override that forgets to stamp metadata now fails loudly on the driver.
- [`91e6493`](https://github.com/The-AI-Alliance/cube-standard/commit/91e64931f01ec1aebc4a0ac45a0b88e62167da8a) `docs`: call out the `AbstractSeedGenerator` → `TypedBaseModel` breaking change in the canonical benchmark spec, the archived deltas, and the PR description.
- [`0444336`](https://github.com/The-AI-Alliance/cube-standard/commit/0444336cdc4d4691df2eaa15a6b3eaa4e5cb5e96) `docs(benchmark)`: inline comment at `CompositeBenchmark.spawn`'s `task_config.make()` call explaining why we bypass `sub_bench.spawn`.
- [`05186b8`](https://github.com/The-AI-Alliance/cube-standard/commit/05186b8a14fde4d50f7ad1dd28b35b76e5d05c39) `docs(server)`: note in `_load_runtime_context._rehydrate` that the recursion is bounded by the JSON payload (cycles already rejected by `_dump_`'s `json.dumps`).
- [`3a196e4`](https://github.com/The-AI-Alliance/cube-standard/commit/3a196e4cc44f2b4a83f37cee69ef8d7f5081ed95) `test`: add `BenchmarkConfig → JSON → worker` seed-generator pipeline test in `test_seed.py` (custom `AbstractSeedGenerator` subclass with a subclass-specific field round-trips and re-emits identical TaskConfigs); extend `_dump_runtime_context` rejection coverage in `test_server.py` to builtin non-JSON types (set, datetime, function) and to nested non-JSON values.

`make test` still green: 509 passed, 5 skipped.

NicolasAG:
## Follow-up: structurally remove the metadata-stamping concern (review [item 1 from previous review](https://github.com/The-AI-Alliance/cube-standard/pull/118#issuecomment-4329291485))

Pushed [`c6244d4`](https://github.com/The-AI-Alliance/cube-standard/commit/c6244d4) to address [review item #1](https://github.com/The-AI-Alliance/cube-standard/pull/118#issuecomment-4329291485) (metadata stamping has no enforcement).

### What changed

`TaskConfig.task_id` is no longer a field — it's a `@property` derived from `metadata.id`, prefixed with `sub_bench_name` when set (composite routing):

```python
class TaskConfig(ABC, TypedBaseModel):
    metadata: SerializeAsAny[TaskMetadata] = Field(...)
    seed: int | None = None
    tool_config: SerializeAsAny[ToolConfig] | None = None
    sub_bench_name: str | None = None

    @property
    def task_id(self) -> str:
        if self.sub_bench_name is not None:
            return f"{self.sub_bench_name}/{self.metadata.id}"
        return self.metadata.id
```

The `_check_metadata_stamp` validator I added in `fc1a8ed` is gone. With a single source of truth on `metadata`, the drift it was guarding is structurally impossible — there is nothing to mismatch.

### Why this resolves the review concern

The original critique was that a custom `get_task_configs()` override that forgets to stamp metadata produces a confusing worker-side `field required` error rather than a helpful driver-side message. With this refactor:

- `metadata: TaskMetadata = Field(...)` is still required, so Pydantic raises `field required` **at the `yield` site** in the override, on the driver, before any worker sees the config. Same error class as before, but now it surfaces immediately at construction — the override can't yield a malformed config and have it propagate.
- There is no second id field to forget about. The `task_id` / `metadata.id` divergence the validator guarded against is unrepresentable.
- The validator and its custom error message are deleted, replaced by Pydantic's standard required-field machinery. Smaller surface, same guarantee.

### Other side effects

- `BenchmarkConfig.get_task_configs()` no longer passes `task_id=` to the constructor.
- `CompositeBenchmarkConfig.get_task_configs()` only updates `sub_bench_name` on each emitted config — the prefix appears automatically via the property.
- Wire format: `model_dump()` no longer includes a top-level `task_id`. No backward compat with previously stored configs (none exist in production yet).

### Downstream cubes

Six cube-harness cubes (`osworld`, `swebench-verified`, `swebench-live`, `terminalbench`, `webarena-verified`, `miniwob`) override `get_task_configs()` and explicitly pass `task_id=tm.id`. They will fail to construct TaskConfigs against this branch until each drops that line. Tracked as a follow-up PR in cube-harness; left out of this one to keep diffs scoped per repo.

### Validation

`make test`: 509 passed, 5 skipped. `examples/counter-cube` tests: 18 passed. `make lint-check`: clean.

recursix (approved):
reviewed delta.md

PR #120, opened 2026-04-24
fix(cli): cube test — skip throughput by default, per-task progress, fix cleanup noise

recursix:
## Summary

- `cube test` now runs once (compliance only) and shows per-task progress as it goes. Add `--stress` to also measure throughput at 1/2/4 workers.
- Playwright cleanup noise suppressed: "Event loop is closed" during shutdown is expected, not a leak
- `check_reset_reproducibility` creates tasks in separate threads so each thread owns its own event loop — cubes using Playwright sync API no longer conflict on a shared event loop
- Task results table: fixed column headers truncating for long task IDs (min_width + ratio + expand=True)
- New `on_episode_start` / `on_episode_done` callbacks on `run_debug_suite` for live progress

## Before / After

**Before:** runs 4× (compliance + warm + 2-worker + 4-worker), no per-task output, Playwright error spam, broken `reset_reproducibility` for browser cubes

**After:**
```
Running debug suite: workarena_cube.debug
  → workarena.servicenow.multi-chart-value-retrieval
  ✓ workarena.servicenow.multi-chart-value-retrieval  41.4s
  → workarena.servicenow.multi-chart-min-max-retrieval
  ✓ workarena.servicenow.multi-chart-min-max-retrieval  50.7s
```
Throughput section shows: `Skipped in test mode. Use cube test --stress to measure.`

## Test plan

- [ ] `cube test counter-cube` runs once, shows per-task progress, no throughput
- [ ] `cube test counter-cube --stress` runs 4× with throughput table
- [ ] `cube test workarena-cube` no Playwright error spam, reset_reproducibility shows real diff instead of asyncio crash

### DIFF
--- a/cube-resources/cube-browser-playwright/src/cube_browser_playwright/playwright_session.py
+++ b/cube-resources/cube-browser-playwright/src/cube_browser_playwright/playwright_session.py
@@ -169,12 +169,13 @@ class PlaywrightSession(BrowserSession):
         try:
             self._context.close()
         except Exception as e:
-            logger.error("Error closing browser context; resources may be leaked: %s", e)
+            # "Event loop is closed" is expected when stop() is called during interpreter
+            # shutdown or after the sync-API event loop has already been torn down — not a leak.
+            if "Event loop is closed" not in str(e):
+                logger.error("Error closing browser context; resources may be leaked: %s", e)
         try:
             self._playwright.stop()
         except Exception as e:
-            logger.error("Error stopping playwright; process may be leaked: %s", e)
-        try:
-            shutil.rmtree(self._user_data_dir)
-        except OSError as e:
-            logger.warning("Failed to remove temp profile dir %r: %s", self._user_data_dir, e)
+            if "Event loop is closed" not in str(e):
+                logger.error("Error stopping playwright; process may be leaked: %s", e)
+        shutil.rmtree(self._user_data_dir, ignore_errors=True)
--- a/openspec/specs/cli/spec.md
+++ b/openspec/specs/cli/spec.md
@@ -35,6 +35,8 @@ Exits non-zero if any debug task fails to reach `reward == 1.0`.
 
 Options:
 - `--max-steps N` — per-episode step budget (default 20)
+- `--stress N` — run each task N times to surface flakiness (default 1)
+- `--no-reset-check` — skip reset reproducibility check (useful in CI where the check is slow)
 - `--output PATH` — save report JSON
 - `--ci` — suppress Rich dashboard, plain-text output (also enabled by `CUBE_CI=1`)
 
--- a/openspec/specs/testing/spec.md
+++ b/openspec/specs/testing/spec.md
@@ -46,17 +46,20 @@ Report schema:
 }
 ```
 
-### `run_debug_suite(benchmark_name, module, *, max_steps=20, workers=1) -> list[dict]`
+### `run_debug_suite(benchmark_name, module, *, max_steps=20, workers=0, on_episode_start=None, on_episode_done=None) -> list[dict]`
 Discovers tasks via `module.get_debug_benchmark().get_task_configs()`, runs each
 with `module.make_debug_agent(task_id)`, and returns a list of episode reports.
 
-**Parallel runs (`workers > 1`):** tasks share the benchmark's `_runtime_context`
-by reference. After `setup()` returns, concurrent episodes must treat that object
-as read-only. Writes from multiple workers are not safe.
+`workers=0` (default) runs one thread per task automatically. `workers=1` is sequential.
+
+**Parallel runs (`workers=0` or `workers > 1`):** tasks share the benchmark's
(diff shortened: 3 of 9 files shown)


NicolasAG:
### Code review

Found 3 issues:

1. `assert_debug_tasks_reward_one` silently switches to parallel execution — breaking change for existing benchmarks.

   The PR changes `run_debug_suite`'s `workers` default from `1` to `0` (auto: one thread per task). `assert_debug_tasks_reward_one` calls `run_debug_suite` without an explicit `workers=` argument, so all existing benchmarks using this helper will silently run all tasks in parallel after this PR. The module docstring's thread-safety warning (`workers > 1`) is now triggered at the wrong threshold. This could cause non-deterministic failures in benchmarks that were written assuming sequential execution and may share mutable state in `_runtime_context`.

   https://github.com/The-AI-Alliance/cube-standard/blob/d873664d0d1281e313a3d445cde04812e353f97f/src/cube/testing.py#L633-L637

2. Spec files not updated for new API surface ([CLAUDE.md](https://github.com/The-AI-Alliance/cube-standard/blob/d873664d0d1281e313a3d445cde04812e353f97f/CLAUDE.md) says "For substantive changes to a spec's contract, write a delta spec in `openspec/changes/<name>/deltas.md` first" and "For completed changes... apply deltas to the main spec.")

   `openspec/specs/testing/spec.md` still documents `workers=1` as the default and has no mention of the new `on_episode_start`/`on_episode_done` callbacks. `openspec/specs/cli/spec.md` does not list the new `--stress` and `--no-reset-check` flags.

   https://github.com/The-AI-Alliance/cube-standard/blob/d873664d0d1281e313a3d445cde04812e353f97f/openspec/specs/testing/spec.md#L47-L52

   https://github.com/The-AI-Alliance/cube-standard/blob/d873664d0d1281e313a3d445cde04812e353f97f/openspec/specs/cli/spec.md#L34-L41

3. Dead no-op self-assignment in `_on_start` callback — `_total_count[0] = _total_count[0]` does nothing. `_total_count` is never updated anywhere, making both the assignment and the variable vestigial. The comment "populated after first call; use running index" suggests an increment (`+= 1`) was intended but never written.

   https://github.com/The-AI-Alliance/cube-standard/blob/d873664d0d1281e313a3d445cde04812e353f97f/src/cube/cli.py#L490-L494

Minor issues (non-blocking):

4. Module docstring in `testing.py` still shows `workers=1` and the thread-safety warning says "Parallel runs (`workers > 1`)" — both are now wrong. The first thing benchmark authors see is the stale signature.

   https://github.com/The-AI-Alliance/cube-standard/blob/d873664d0d1281e313a3d445cde04812e353f97f/src/cube/testing.py#L5-L26

5. PR description says `check_reset_reproducibility` "creates tasks sequentially (close t1 before making t2)" but the implementation uses `ThreadPoolExecutor(max_workers=2)` with both futures submitted simultaneously — they run in parallel threads. The in-code docstring is correct; the PR summary is not.

6. `assert_debug_tasks_reward_one` docstring describes it as "a drop-in for pytest" with no mention that it now runs all tasks in parallel. A short note about the `_runtime_context` read-only requirement (mirroring the module-level warning) would prevent surprises for benchmark authors.

   https://github.com/The-AI-Alliance/cube-standard/blob/d873664d0d1281e313a3d445cde04812e353f97f/src/cube/testing.py#L608-L637

recursix:
Thanks for the thorough review. Here's a status on each point:

**Issues 2, 3, 4 — fixed in this branch:**
- Spec files updated: `openspec/specs/testing/spec.md` now shows `workers=0` default and documents `on_episode_start`/`on_episode_done`; `openspec/specs/cli/spec.md` lists `--stress` and `--no-reset-check`.
- Dead self-assignment fixed: `_total_count[0] = _total_count[0]` → `_total_count[0] += 1`.
- Module docstring updated to `workers=0` with the correct thread-safety threshold.

**Issue 5 (minor) — noted; PR description was wrong, in-code docstring is correct.**

**Issue 1 — intentional, now with a detection mechanism:**

The switch to `workers=0` (auto-parallel) in `assert_debug_tasks_reward_one` is intentional. Any benchmark that breaks under parallel execution has a latent thread-safety bug that exists independently of this PR — surfacing it sooner is better than hiding it behind sequential execution. The correct fix for such a benchmark is to make it thread-safe, not to revert the default here.

That said, silent failures are hard to debug, so we added a cheap post-run detection mechanism: `run_debug_suite` now takes an `id()` snapshot of `benchmark._runtime_context` before and after parallel execution and logs a `WARNING` if any key is added, removed, or reassigned. We use `id()` rather than `deepcopy` intentionally — deepcopy fails on non-serializable objects like browser connections.

The `assert_debug_tasks_reward_one` docstring now explicitly documents the `_runtime_context` read-only requirement.

**Issue 6 (minor) — fixed as part of the above.**

NicolasAG (approved):

Issue #119, opened 2026-04-23
Benchmark's default_tool_config is not default but designated


erikchwang:
Benchmark defines `default_tool_config: ToolConfig` as a field, and describe it as `default tool config to be used for tasks that don't specify their own`. 

This means each task is allowed to have its customized tool. 

However, in `Benchmark.get_task_configs()`, task configs are always created with `default_tool_config`: 

https://github.com/The-AI-Alliance/cube-standard/blob/144ffc9392b88278c1c74c9c35c7ad64d26dd384/src/cube/benchmark.py#L440

So each task cannot have its customized tool. 

This is not the purpose of defining `default_tool_config`. 

A way to improve is:

1. Add `tool_config: ToolConfig` in `TaskMetadata` as an optional field. 

2. In `Benchmark.get_task_configs()`, for each `tm`, use `tm.tool_config` to create the task config, while using `default_tool_config` only if `tm.tool_config` is `None`.

NicolasAG:
Thanks for raising this! You're right that the name is misleading.

The base `get_task_configs()` doesn't actually implement any fallback; calling the field `default_tool_config` suggests semantics the framework doesn't provide. That's the real bug. I'll rename it to `tool_config` and rewrite the docstring to say plainly: this is the tool config applied uniformly to every task; per-task variation is the benchmark author's responsibility, via either:

- overriding `get_task_configs()` (what every shipped cube already does: https://github.com/The-AI-Alliance/cube-harness/tree/main/cubes), or
- constructing `TaskConfig` objects directly, bypassing `get_task_configs()` entirely.

I'd push back on putting `tool_config` inside `TaskMetadata`, though. `TaskMetadata` is the static description of what a task _is_, sent to parallel workers and shipped as a package resource. A `ToolConfig` is an _interface choice_ (Playwright vs. Selenium, Computer vs. accessibility tree) that's the experiment's concern, not the task's. Embedding it in metadata couples shipped JSON to specific tool packages and conflates "what the task needs" with "how the agent interacts." That's why `tool_config` lives on `TaskConfig` (per-run) rather than `TaskMetadata` (per-task).

I'll open a small PR for the rename + docs.

erikchwang:
solved by renaming: https://github.com/The-AI-Alliance/cube-standard/pull/123

erikchwang:
[...]

PR #121, opened 2026-04-25
feat: typed TaskExecutionInfo slot + remove extra_info bags

NicolasAG:
> Stacks on top of #118. Target branch is `feat/benchmark-config`, not `main`.

## Summary

Follow-up to #118 that resolves the **\"`extra_info` split\"** open design question listed in that PR's description. Introduces a typed `Task.execution_info` slot for heavy per-task data, deletes the stringly-typed `extra_info: dict[str, Any]` bags on both `TaskMetadata` and `BenchmarkMetadata`, and relocates the per-task cache helpers so workers never need to import the owning `BenchmarkConfig` (preserving #118's worker-side decoupling).

Implementation now lands across 7 follow-up commits on this branch (linked in the table below).

## Problem & constraints

The redesign was driven by nine concrete constraints, all observed in the existing cubes (`osworld-cube`, `swebench-live-cube`, `swebench-verified-cube`, `terminalbench-cube`). Any acceptable design had to satisfy **all** of them; that's what ruled out the alternatives below.

1. **Stringly-typed access is harmful.** `metadata.extra_info["instruction"]` has no autocomplete, no schema validation, and fails at runtime on typos. The dict-of-Any shape encourages sloppy use and makes refactoring fragile.
2. **`TaskMetadata` must stay lean.** It powers `cube list`, registry listings, filtering, glob-based subsetting, and human inspection. It ships in the wheel and is loaded eagerly. Heavy data has no business being in it.
3. **Eager and lazy data have different lifetimes — the type system should reflect that.** `TaskMetadata` lives in the wheel from class definition. Heavy execution data lives on the worker, loaded per-task at construction or reset. Today's pattern smuggles the second through the shape of the first; the type system can't distinguish them.
4. **`install()` does real preprocessing, not just downloading.** OSWorld clones the upstream repo, extracts per-task evaluator configs, normalizes machine-specific paths, and writes processed JSON to a cache. This processed-cache layer is load-bearing — replacing it with "just call HF Datasets at task time" would lose the work `install()` does.
5. **Workers may run on different machines than the one that ran `install()`.** Local-disk caches don't cross machines. Any design must work when driver and workers are entirely separate hosts (typical at scale).
6. **PR #118's invariant must hold: workers MUST NOT import the owning `BenchmarkConfig`.** That was the whole point of #118's `TaskConfig` self-containment refactor. Any new cache-resolution path that requires `type(self).benchmark_class.<something>` on the worker side regresses #118.
7. **Concurrency at scale rules out per-task downloads.** With ≥hundreds of workers, having every worker download/process data on first task call is unacceptable. The setup must be a one-time phase per worker environment, not a per-task operation.
8. **Heterogeneous data sources resist a single abstraction.** swebench data lives in HuggingFace Datasets. terminalbench data lives in a git repo. OSWorld pre-processes via a custom upstream clone. Future cubes may use S3, internal APIs, or generated data. A single framework data-distribution abstraction would either be too narrow (favoring one source) or too broad (becoming a new mini-platform that competes with mature ecosystem tools — HF, git, Docker images, EFS).
9. **Existing cubes use an undocumented workaround.** All three production cubes today merge `load_task_execution_info()` into `metadata.extra_info` via `model_copy(update=...)` inside `make()`. It works, but it's not in the spec, reviewers can't tell at a glance whether a given `extra_info` value is a lightweight tag or a heavy payload, and `review-cube` has no way to flag eager-loaded heavy data on `TaskMetadata`.

The combination is what's hard. Constraint (2) rules out leaving heavy data on `TaskMetadata`. (4) rules out deleting the processed-cache layer. (5) rules out laptop-style local-only caches. (6) rules out resolving cache paths on the worker via `BenchmarkConfig`. (7) rules out lazy-fetch-on-task-call without a one-time setup. (8) rules out a framework-owned distribution layer. (9) means whatever replaces `extra_info` must have teeth in `review-cube`.

## Background — what we considered and rejected

The conversation that produced this design walked through a ladder of options. Recorded here so we don't re-derive it next time:

### Option A — Add a third metadata class (light + heavy)

Rejected. The lifetime difference (eager vs. lazy) is real, but adding a third typed class alongside `TaskMetadata` doubles the surface without addressing the underlying problem (stringly-typed access via `metadata.extra_info[\"...\"]`). Heavy data clearly belongs on the **`Task`**, not on a sibling of `TaskMetadata`.

### Option B — Build a pluggable `DataStore` / `BlobRef` abstraction

Rejected. A clean, well-typed framework component to manage content-addressed heavy-data distribution across workers (local disk / S3 / HF / git) — **and exactly the kind of platform-thinking #118 was built to avoid.**

The framework's job is the **protocol** (gym-style step/reset/evaluate, JSON-RPC boundary, `BenchmarkConfig.make()`), not the **plumbing** (where bytes live, how they get to workers). Heavy-data distribution is solved by mature ecosystem tools — HuggingFace Datasets, git, Docker images, EFS — that operators already use. cube-standard absorbing that responsibility would invent infrastructure that already exists.

### Option C — Move heavy data into `TaskConfig` and ship it inline

Rejected. Defeats the JSON-RPC boundary that #118 just established. Each `TaskConfig` would balloon to MB for swebench-shape data, and N-task batches would carry N copies over the wire. The whole point of the worker boundary is that it stays cheap.

### Option D — Centralize via a registry resolver hook (`benchmark_ref`)

Deferred (already documented in #118's [10x-vs-100x comment](https://github.com/The-AI-Alliance/cube-standard/pull/118#issuecomment-4305165062)). Compatible with this PR's design — if a registry layer ever becomes load-bearing, `benchmark_ref` is the hook point. Not building it now.

### What we landed on (this PR)

Smallest principled change:

1. **Typed slot:** `Task.execution_info: TaskExecutionInfo | None`. Cube authors subclass `TaskExecutionInfo` with named fields; populate inside `TaskConfig.make()` (most common) or `Task.model_post_init` / `Task.reset()`.
2. **Delete `extra_info` bags** on `TaskMetadata` and `BenchmarkMetadata`. Force named, typed subclassing for any extra fields. `TaskMetadata` stays lean and human-readable, as intended.
3. **Move `task_execution_cache_dir()` and `load_task_execution_info()` from `BenchmarkConfig` to `TaskConfig`.** Single source of truth for the cache path; no possibility of drift; workers don't import the owning `BenchmarkConfig`. `BenchmarkConfig.install()` writes via `cls.task_config_class.task_execution_cache_dir()`.
4. **Add `TaskConfig.verify_installed()`.** Optional classmethod called by convention from `make()` so misconfigured workers fail fast with an actionable error.
5. **Add `cube install <bench>` CLI subcommand.** Small UX improvement mirroring `cube test <bench>` — operators wire it into Dockerfiles and init scripts.

The framework provides the **typed slot + the install/verify hooks**. Cube authors own **what's in the slot and how it's hydrated** (HF Datasets / git / S3 / local processing — their choice). Operators own **how the cache reaches workers** (baked into image / shared volume / worker bootstrap).

## How the design satisfies each constraint

| # | Constraint | How the design satisfies it |
|---|---|---|
| 1 | Stringly-typed access is harmful | `TaskExecutionInfo` is a typed `TypedBaseModel` subclass. `self.execution_info.instruction` is autocomplete-friendly, schema-validated by Pydantic, and refactor-safe. No more `metadata.extra_info["..."]`. |
| 2 | `TaskMetadata` must stay lean | Deleting `TaskMetadata.extra_info` removes the smuggling channel. Authors needing extras subclass `TaskMetadata` with typed fields. Heavy data lives on `TaskExecutionInfo`, never on `TaskMetadata` — `cube list`, registry tooling, and filtering never touch it. |
| 3 | Lifetime difference must be in the type system | Two distinct types in two distinct slots on `Task`: `metadata: TaskMetadata` (eager, ships in the wheel) and `execution_info: TaskExecutionInfo \| None` (lazy, populated on the worker). The eager/lazy split is explicit at the type level — impossible to accidentally eager-load heavy data because there's no slot for it on `TaskMetadata`. |
| 4 | `install()` does real preprocessing | Contract preserved. `BenchmarkConfig.install()` still downloads, processes, and writes processed JSON to the cache — the only change is it writes via `cls.task_config_class.task_execution_cache_dir()` (single owner of the path). OSWorld's clone-and-normalize pipeline is unchanged. |
| 5 | Workers on different machines | Documented operator deployment model with three concrete strategies (image / shared volume / worker bootstrap). The framework provides the hook; operators choose distribution using mature ecosystem tools (Docker layers, EFS / NFS, init containers). |
| 6 | Workers MUST NOT import `BenchmarkConfig` | `task_execution_cache_dir()`, `load_task_execution_info()`, and `verify_installed()` all live on `TaskConfig`. Worker-side code reads via `type(self).<method>` — never crosses the worker → owning-config boundary. PR #118's invariant holds. |
| 7 | Concurrency at scale | `install()` is a one-time, per-worker-environment phase, never invoked by `make()`. Workers only do fast cache reads. All three operator strategies amortize the cost: image (zero runtime cost, paid at build), shared volume (one-time per cluster), bootstrap (one-time per worker). |
| 8 | Heterogeneous data sources | The framework intentionally provides **no** data-distribution abstraction. `install()` is opaque — what runs inside is the cube author's choice (HF Datasets, git clone, S3 fetch, custom API, generated). Future cubes are not forced through a framework-owned store. |
| 9 | Existing workaround was undocumented | `review-cube` gains rules with teeth: `TaskMetadata` subclasses MUST NOT define `extra_info`; heavy per-task data MUST live on a `TaskExecutionInfo` subclass. The pattern moves from undocumented workaround to typed first-class slot, enforced at review time. |

## Operator deployment model

`install()` runs **once per worker environment**, not once per experiment. Three standard strategies (documented in the spec, not enforced by the framework):

1. **Bake into the worker image** (default for production fleets) — `RUN cube install <bench>` in the Dockerfile.
2. **Shared volume** — mount NFS / EFS / S3FS; run `install()` once on any node.
3. **Worker bootstrap** — each worker runs `cube install <bench>` at startup. Fine for ≤dozens of workers; push to (1) at thousands.

The framework does **not** coordinate downloads across workers, manage a shared cache, or move bytes between machines.

## Concrete invariants this PR enforces

- `TaskMetadata` and `BenchmarkMetadata` no longer accept arbitrary `dict[str, Any]` payloads.
- `TaskConfig` is the single owner of the per-task execution cache directory.
- Workers never import the owning `BenchmarkConfig` to construct a `Task` — including for cache verification (the `verify_installed()` hook lives on `TaskConfig`).
- `subset_from_glob` no longer special-cases `\"extra_info.<key>\"` dot-notation; it operates on top-level fields of typed `TaskMetadata` subclasses.
- CSV/JSON loaders no longer have an `extra_info` JSON column convention.

## What's in this PR

| Commit | Scope |
|---|---|
| [`a110073`](https://github.com/The-AI-Alliance/cube-standard/pull/121/commits/a110073) | OpenSpec proposal + deltas folder (`typed-task-execution-info/`) |
| [`58641ab`](https://github.com/The-AI-Alliance/cube-standard/pull/121/commits/58641ab) | `cube.task` — `TaskExecutionInfo` base class, `Task.execution_info` field, relocated cache helpers (`task_execution_cache_dir`, `load_task_execution_info`, `verify_installed`) on `TaskConfig`. Drops `TaskMetadata.extra_info`. |
| [`5238b9b`](https://github.com/The-AI-Alliance/cube-standard/pull/121/commits/5238b9b) | `cube.benchmark` — drops `BenchmarkMetadata.extra_info`, drops relocated cache helpers from `BenchmarkConfig`, simplifies `subset_from_glob` (no more `extra_info.<key>` dot-notation), drops the CSV `extra_info` JSON column convention. |
| [`a9e7a1f`](https://github.com/The-AI-Alliance/cube-standard/pull/121/commits/a9e7a1f) | `cube.cli` — `cube install <bench>` subcommand + `test_cli` coverage. |
| [`57ae050`](https://github.com/The-AI-Alliance/cube-standard/pull/121/commits/57ae050) | `_template/` scaffold — commented `TaskExecutionInfo` stub, `verify_installed()` override stub, install-pattern hook in `benchmark.py`, no `extra_info` anywhere. |
| [`a88ad26`](https://github.com/The-AI-Alliance/cube-standard/pull/121/commits/a88ad26) | counter-cube + toy_benchmark + toy_bench_auto_load_from_json migrate to typed `CounterTaskMetadata` (typed `target` / `difficulty` / `tool_overrides`). The auto-load JSON example uses `_type` discriminator to instantiate the typed subclass. Two pre-existing test invocations get the missing `metadata=` kwarg. |
| [`61b1a28`](https://github.com/The-AI-Alliance/cube-standard/pull/121/commits/61b1a28) | `.claude/skills/{new-cube,review-cube}` — interview, scaffold, and review checks reflect the typed shape; `review-cube` blocks any `TaskMetadata` subclass that declares `extra_info` (or any `dict[str, Any]` bag) and any heavy data on `TaskMetadata` instead of `TaskExecutionInfo`. |
| [`5ee3f74`](https://github.com/The-AI-Alliance/cube-standard/pull/121/commits/5ee3f74) | OpenSpec archive: moves `typed-task-execution-info/` → `archive/2026-04-27-typed-task-execution-info/` and applies deltas to `task/spec.md`, `benchmark/spec.md`, `cli/spec.md`. |

## Out of scope

- **Cube migration** (osworld-cube, swebench-live-cube, swebench-verified-cube, terminalbench-cube). Those land as parallel PRs on **cube-harness**. The contract is mechanical: drop the `model_copy(update={\"extra_info\": ...})` dance, replace `metadata.extra_info[\"x\"]` with `execution_info.x`, point `install()` writes at `cls.task_config_class.task_execution_cache_dir()`.
- Any data-distribution layer (rejected — Option B above).
- Backwards-compat shims for `extra_info`. Per the project's no-shims policy, cubes migrate in parallel.
- `BenchmarkPool` (still deferred from #118).

## Test plan

- [x] `make test` passes after each follow-up commit (365 passed, 5 skipped)
- [x] `make lint` clean (pre-commit hooks pass on every commit)
- [x] OpenSpec deltas applied to `task/spec.md`, `benchmark/spec.md`, `cli/spec.md`
- [x] `_template/` scaffold builds and the package imports + emits TaskConfigs cleanly
- [x] counter-cube tests pass (18/18); `examples/toy_benchmark/counter.py` and `examples/toy_bench_auto_load_from_json/counter.py` run end-to-end
- [ ] `new-cube` and `review-cube` skills dry-run-tested (left for a manual pass before merge)
- [x] OpenSpec change archived to `archive/2026-04-27-typed-task-execution-info/`


NicolasAG:
### Code review

Found 7 blocking issues (all fixed) + 4 minor issues noted below.

---

#### Blocking issues

1. **`verify_installed()` template crashes with `FileNotFoundError` instead of `RuntimeError`** — `Path.iterdir()` raises when the cache dir doesn't exist (the not-installed state it's meant to catch). Needs an `exists()` guard before the empty check. → Fixed in [f47a818](https://github.com/The-AI-Alliance/cube-standard/commit/f47a818934e98c96479991debc39ccf9c472de19)

https://github.com/The-AI-Alliance/cube-standard/blob/7eb72801d48f929c974b9b4f02b9d99be665c49a/src/cube/_template/new_cube_package/src/cube_package/task.py#L86-L98

2. **`cube install` will `TypeError` for `CompositeBenchmarkConfig`** — `cmd_install` calls `cls.install()` on the entry-point class, but `CompositeBenchmarkConfig.install` is intentionally an instance method (sub-configs are instance state). A composite registered as a `cube.benchmarks` entry point will crash with `install() missing 1 required positional argument: 'self'`. → Fixed in [41c7321](https://github.com/The-AI-Alliance/cube-standard/commit/41c7321557857e647787b73a9d3d41827b192f69): CLI now detects composites and exits 1 with a clear message directing the user to install each sub-benchmark individually.

https://github.com/The-AI-Alliance/cube-standard/blob/7eb72801d48f929c974b9b4f02b9d99be665c49a/src/cube/benchmark.py#L893-L901

https://github.com/The-AI-Alliance/cube-standard/blob/7eb72801d48f929c974b9b4f02b9d99be665c49a/src/cube/cli.py#L402-L414

3. **`cmd_install` interpolates exception text into Rich markup without escape** — any exception whose message contains `[…]` (paths, tag-like text) raises `rich.errors.MarkupError` and hides the real failure. `cmd_test` already uses `rich.markup.escape`; this should match. → Fixed in [4bb1705](https://github.com/The-AI-Alliance/cube-standard/commit/4bb170565edd9a05ca1fa59dff4a90891f8be670)

https://github.com/The-AI-Alliance/cube-standard/blob/7eb72801d48f929c974b9b4f02b9d99be665c49a/src/cube/cli.py#L405-L414

4. **Cache path formula changed silently** — moved from `~/.cube/<benchmark_metadata.name>/…` to `~/.cube/<top-level-python-package>/…` (e.g. `osworld-cube` → `osworld_cube`). Existing operator caches become invisible after upgrade with no migration note in the PR description or archived proposal. → Documented in [2637a5c](https://github.com/The-AI-Alliance/cube-standard/commit/2637a5c7949b9514475665f89b831a35dec23d55) (no backward compat shim needed — no external users yet).

https://github.com/The-AI-Alliance/cube-standard/blob/7eb72801d48f929c974b9b4f02b9d99be665c49a/src/cube/task.py#L487-L500

5. **Template `benchmark_metadata.json` still ships `"extra_info": {}`** — `task_metadata.json` was cleaned up but the sibling file wasn't. Cubes scaffolded by `cube init` get a stale field that Pydantic silently drops. → Fixed in [347a425](https://github.com/The-AI-Alliance/cube-standard/commit/347a4254b35ab16349f44adb49028d94c8f46887)

https://github.com/The-AI-Alliance/cube-standard/blob/7eb72801d48f929c974b9b4f02b9d99be665c49a/src/cube/_template/new_cube_package/src/cube_package/benchmark_metadata.json#L1-L13

6. **Spec drops `SerializeAsAny` from `Task.execution_info`** — code uses `SerializeAsAny[TaskExecutionInfo]` (load-bearing for the typed-execution-info contract), but the spec shows the bare `TaskExecutionInfo | None`. A cube author following the spec literally loses subclass-specific fields on JSON round-trip. → Fixed in [44e38a9](https://github.com/The-AI-Alliance/cube-standard/commit/44e38a990212fef54ad225468a070ddda17d91ae)

https://github.com/The-AI-Alliance/cube-standard/blob/7eb72801d48f929c974b9b4f02b9d99be665c49a/openspec/specs/task/spec.md#L66-L75

7. **Archived proposal still says `Status: Proposed`** — sibling archive `2026-04-24-benchmark-config/proposal.md` uses `Accepted`. Inconsistent with the archive workflow described in [CLAUDE.md](https://github.com/The-AI-Alliance/cube-standard/blob/7eb72801d48f929c974b9b4f02b9d99be665c49a/CLAUDE.md#L26-L34). → Fixed in [97795c4](https://github.com/The-AI-Alliance/cube-standard/commit/97795c4963b5a0dfab40fc2749738ba97dd721a9)

https://github.com/The-AI-Alliance/cube-standard/blob/7eb72801d48f929c974b9b4f02b9d99be665c49a/openspec/changes/archive/2026-04-27-typed-task-execution-info/proposal.md#L1-L8

---

#### Minor issues (post-review, not blocking)

8. **`tool_overrides` on `CounterTaskMetadata` violates design principles** — tool config is a runner concern, not task metadata (see [#119](https://github.com/The-AI-Alliance/cube-standard/issues/119)). The counter-cube and toy examples stored `tool_overrides: CounterToolConfig | None` on metadata and resolved it in `make()`, conflating "what a task is" with "how to run it". → Fixed in [63b2382](https://github.com/The-AI-Alliance/cube-standard/commit/63b2382f53d552460d9325631bfc6eab55da190c): replaced with a `_TASK_TOOL_CONFIGS` ClassVar lookup dict on `BenchmarkConfig`, resolved in `get_task_configs()`.

9. **`subset_from_glob` dot-notation removed without a migration error** — the old `"extra_info.<key>"` glob syntax now silently returns an empty subset (Python `hasattr` returns `False` for dotted attribute names). Not blocking since no external users exist yet and no backward compat shim is needed per project policy.

10. **CSV loader `_type` dispatch** — `task_metadata_from_csv` calls `TaskMetadata.model_validate(row)` per row; if a CSV had a `_type` column the dispatch would work (handled by `TypedBaseModel`'s `_deserialize_with_type` validator), but CSVs don't naturally carry `_type`. Pre-existing limitation, not introduced by this PR.

11. **`CounterTaskConfig.make()` missing `metadata: CounterTaskMetadata` narrowing annotation** — `make()` accesses typed fields on metadata but the annotation is the base `TaskMetadata`. Low risk since `tool_overrides` access was removed in issue 8; only a typechecker concern.

---

recursix (approved):

PR #123, opened 2026-04-28
fix(benchmark): rename default_tool_config → tool_config

### LINKED ISSUE #119: Benchmark's default_tool_config is not default but designated

NicolasAG:
## Summary

- Renames `Benchmark.default_tool_config` → `tool_config` across `benchmark.py`, the benchmark spec, and the architecture diagram
- The `default_` prefix implied fallback semantics (used only when a task has no tool of its own), but the base `get_task_configs()` always applies it uniformly to every task — the name was misleading
- Updates the docstring and spec to say what it actually does: applied to every task by the default `get_task_configs()`; override `get_task_configs()` for per-task variation

Closes #119.

## Test plan

- [x] `make test` — 467 passed, 5 skipped

### DIFF
--- a/architecture-diagram.md
+++ b/architecture-diagram.md
@@ -166,7 +166,7 @@ classDiagram
         +ClassVar~type~ task_config_class
         -RuntimeContext _runtime_context
         +ContainerBackend container_backend
-        +ToolConfig default_tool_config
+        +ToolConfig tool_config
         +AbstractSeedGenerator seed_generator
         +setup() void
         +get_task_configs() Generator~TaskConfig~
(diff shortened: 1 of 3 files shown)


erikchwang:
Thanks for the renaming!

erikchwang (approved):
LGTM

PR #124, opened 2026-04-29
Small updates while migrating cubes in cube-harness #322 #323 #324

### LINKED ISSUE #127: _is_dynamic() in BenchmarkConfig.__init_subclass__()
https://github.com/The-AI-Alliance/cube-standard/blob/9982b279dd780bdae95b0f8cbcf2a7f3fab79d0e/src/cube/benchmark.py#L301

The current `any(...)` keeps walking past entries that don't match.

But consider:

```
class A(BenchmarkConfig):
    @property
    def benchmark_metadata(self): ...

class B(A):
    benchmark_metadata: ClassVar[BenchmarkMetadata] = BenchmarkMetadata(...)

class C(B):
    pass
```

When `BenchmarkConfig.__init_subclass__()` runs for `C`, `_is_dynamic()` will return True, but it should be False. 

The fix is: walk the MRO, stop at the first class that defines the name, and return whether it is a property:

```
def _is_dynamic(name: str) -> bool:
    for klass in cls.__mro__:
        if name in klass.__dict__:
            return isinstance(klass.__dict__[name], property)
    return False
```

NicolasAG:
## Summary

A grab-bag of small, independent cube-standard improvements driven by what I hit while
migrating cubes in cube-harness PRs #322 / #323 / #324. None of these change runtime
behaviour for existing cubes; each commit stands alone.

### `03fa2e7` — use `typing.Self`

Cosmetic: replace explicit return-type annotations like `-> "WorkArenaBenchmarkConfig"`
on `model_validator(mode="after")` with `typing.Self`. Mechanical, no behaviour change.

### `647742c` — make `TaskConfig`, `BenchmarkConfig`, `Benchmark` generic over their narrowable types

The bigger one. Promotes three classes to generics so cube subclasses can **opt in** to
narrowed static types without re-annotating fields on the subclass (which is what every
cube has been doing today, with `# type: ignore` to silence the inevitable type-checker
complaint). Reviewer flagged the re-annotation pattern as the central typing smell of
the project — invariant fields can't be soundly narrowed, only covariant accessors can.

| Layer | Today | After this PR |
|---|---|---|
| `TaskConfig` | `class FooTaskConfig(TaskConfig): metadata: SerializeAsAny[FooTaskMetadata]  # type: ignore` | `class FooTaskConfig(TaskConfig[FooTaskMetadata]):` — no override |
| `BenchmarkConfig` | `class FooBenchmarkConfig(BenchmarkConfig): task_metadata: ClassVar[dict[str, FooTaskMetadata]]  # type: ignore` | `class FooBenchmarkConfig(BenchmarkConfig[FooTaskMetadata]):` — narrowing via `cfg.tasks() -> Mapping[str, FooTaskMetadata]` (covariant) |
| `Benchmark` | `class FooBenchmark(Benchmark): config: "FooBenchmarkConfig"  # type: ignore` | `class FooBenchmark(Benchmark["FooBenchmarkConfig"]):` — no override |

OpenSpec proposal + deltas in `openspec/changes/task-config-generic/`.

### `514d6a8` — replace TypeVar/Generic with PEP 695 class-scoped type parameters (#125)

Python ≥3.12 lets us write `class Foo[T: Bound]` instead of declaring module-level
TypeVars and mixing in `Generic`. Converts `TaskConfig`, `BenchmarkConfig`, and
`Benchmark` to the new syntax — drops the `TTMetadata` / `TBenchConfig` declarations
and the `Generic` / `TypeVar` imports.

### `37befa4` → `1039e14` — drop, then restore `BenchmarkConfig[TTMetadata]` generic

Tried to drop the `BenchmarkConfig` generic on the theory it was unused, then reverted
once cube-harness migration showed it wasn't: `MiniWobBenchmarkConfig.get_task_configs`
passes `tm` (typed `TaskMetadata` from a non-generic base) into
`MiniWobTaskConfig(metadata=tm, ...)`, whose `metadata` is narrowed via
`TaskConfig[MiniWobTaskMetadata]` — Pylance rejects the call without a generic on
`BenchmarkConfig`. Net effect of the two commits: no change vs. `647742c`, but the
cast in `tasks()` now has a comment explaining why it's necessary (PEP 526 forbids
`ClassVar[dict[str, TTMetadata]]`, so the registry stays at the base type and the
narrowing happens at the read site).

### `db401de` — format lint

Lint-only.

### `afde103` — make `Task` generic over `TTMetadata`

Extends the same PEP 695 narrowing pattern to `Task` itself. `self.metadata` on a
parametrised `Task[FooTaskMetadata]` subclass is now statically typed as
`FooTaskMetadata` — no more `metadata: FooTaskMetadata  # type: ignore[assignment]`
re-annotation in cubes. Pairs naturally with `TaskConfig[FooTaskMetadata]`.

### `7711c73` — constrain `azure-mgmt-compute` to `<38.0`

Dependency pin (from Aman). Unblocks resolution.

### `81c5ceb` — walk MRO in `_is_dynamic` instead of `any()` (#127, via #130)

Bug fix: `_is_dynamic` used `any(... for klass in cls.__mro__)`, so a parent's
`@property` would mark the attribute dynamic even when a child shadowed it with a
`ClassVar` — skipping the file-load / type-check branch in `__init_subclass__` and
letting invalid `ClassVar` overrides through. Now walks the MRO and decides based on
the *nearest* definition. Closes #127.

## Backwards compatibility

Fully backwards-compatible at runtime. Every existing `class FooTaskConfig(TaskConfig):`,
`class FooBenchmarkConfig(BenchmarkConfig):`, `class FooBenchmark(Benchmark):`, and
`class FooTask(Task):` keeps working unchanged — type parameters resolve to their
bounds for unparametrised subclasses, so static types match what they were before.
Migration to the parametrised forms is opt-in and per-cube, landing in cube-harness
#322 / #323 / #324.

## Test plan

- [x] `make lint` clean
- [x] `make test` — 524 passed, 5 skipped (pre-existing infra skips: Docker / Modal / Daytona / Toolkit / sample data)
- [x] Smoke test: parametrised `TaskConfig[FooTaskMetadata]`, `BenchmarkConfig[FooTaskMetadata]`, `Benchmark["FooBenchmarkConfig"]`, and `Task[FooTaskMetadata]` end-to-end
- [x] Smoke test: unparametrised legacy form still constructs and round-trips identically
- [x] Regression test for the `_is_dynamic` MRO fix (invalid `ClassVar` override on a child of a class with `@property benchmark_metadata` is rejected)
- [ ] Downstream verification in cube-harness PRs #322 / #323 / #324


openspec/changes/archive/2026-05-01-task-config-generic/deltas.md
```diff
@@ -0,0 +1,293 @@
...
+still a `dict`; only the static contract changes.
+
+**Why `task_metadata: ClassVar` is *not* narrowed.** Two independent reasons:
+
+1. PEP 526 (and pyright/mypy in strict mode) forbid TypeVars inside
+   `ClassVar` — a `ClassVar` is shared across all generic specialisations,
+   but TypeVars vary per specialisation, so combining them is incoherent.
+2. Even setting PEP 526 aside, `dict` is invariant in its value type, so
+   `dict[str, FooTaskMetadata]` is not a subtype of `dict[str,
+   TaskMetadata]` — narrowing the registry directly would be unsound for
```
conte91, line 148: FYI The abstract/invariant version is `Mapping[str, TaskMetadata]` (which can be correctly subclassed by `Mapping[str, FooTaskMetadata]`). Because the problem with `dict`s is that any generic object can be appended to them, but `Mapping` being a read-only interfact, the subclassing of the values you read can be guaranteed!

conte91: (Not super relevant to the discussion, but it could be good to update the classes which we can't be bothered converting now to use `Mapping` so the typing is correct without the big operational cost of embracing the "generic" interface, could be a good middleground to solve the problem you're describing)

src/cube/benchmark.py
```diff
@@ -394,18 +424,24 @@ def num_tasks(self) -> int:
...
+        type) so subclasses parametrised as ``BenchmarkConfig[FooTaskMetadata]``
+        get a narrowed view at every read site without needing to override
+        the registry. The runtime value is still a ``dict`` — ``Mapping`` is
+        only the static contract.
         """
         full = self.task_metadata
         if self.task_ids is None:
-            return full
-        return {tid: full[tid] for tid in self.task_ids}
+            return cast(Mapping[str, TTMetadata], full)
```
conte91, line 452: Here and elsewhere: you don't need this if `full` is a `dict[str, TTMetadata]`. python's typing system is smart enough so a plain `return full` should suffice. If it doesn't, there's some other quirk which should be investigated (but I believe it's ok since the previous version was returning `full` with no type errors!)

`cast` is terribly named and not actually do any check - it just silently swallows any type assumption (like `dynamic_cast` does in C++), so if `full` every becomes incompatible with the return type we'll never know :) (basically, `cast` is a `type: ignore` in disguise!)

conte91 (approved):
Thanks, love this! I sent you https://github.com/The-AI-Alliance/cube-standard/pull/125 on top which simplifies the syntax a bit given the requirement for python 3.12

src/cube/benchmark.py
NicolasAG: Screenshot 2026-04-29 at 2 21 23 PM
This is the warning that I get when I remove the `cast`

openspec/changes/archive/2026-05-01-task-config-generic/deltas.md
NicolasAG: it's ok as it is now I think. PEP 526 independently blocks narrowing the ClassVar at the framework level anyway.
Also, the gotcha section in this doc explicitly steers cubes away from re-annotating on subclasses in favour of the Generic[TTMetadata] + tasks() path.

src/cube/benchmark.py
NicolasAG: Actually, this is needed.

PR #128, opened 2026-04-30
fix(infra_local): rate-limit retry + active.json locking + Podman DOCKER_HOST

recursix:
## Summary

Three related fixes to `LocalInfraConfig` / `infra_local.py`:

- **Rate-limit retry**: `LocalContainerBackend` retries Docker Hub 429 errors with exponential backoff — removes the need for pre-pull loops in benchmark setup code
- **`active.json` concurrency fix**: `_register_active` / `_deregister_active` now use `fcntl.flock` + atomic `tmp → replace` writes; concurrent Ray workers at ≥12 parallelism were corrupting the file with simultaneous writes
- **Podman `DOCKER_HOST` normalization**: Podman sets `DOCKER_HOST=http+unix://…` but the Docker Python SDK only accepts `unix://`; strip the `http+` prefix in `LocalDockerServiceHandle` before calling `docker.from_env()`, so every downstream user gets this transparently without recipe-level workarounds

## Follow-up

The `flock` approach is correct but still serializes all workers through one lock. #129 tracks a simpler design: one file per run in `~/.cube/active/<run_id>.json`, which eliminates the lock entirely.

## Test plan

- [x] `active.json` corruption no longer observed at 12+ parallel Ray workers
- [x] Docker Hub rate-limit errors retry automatically instead of crashing
- [x] Podman users can run recipes without normalizing `DOCKER_HOST` themselves

### DIFF
--- a/src/cube/backends/local.py
+++ b/src/cube/backends/local.py
@@ -46,8 +46,10 @@ _retry_pull = retry(
     stop=stop_after_attempt(6),
     wait=wait_exponential(multiplier=30, min=30, max=300),
     retry=retry_if_exception(
-        lambda exc: isinstance(exc, docker.errors.APIError)
-        and any(kw in str(exc).lower() for kw in ("toomanyrequests", "rate limit", "429"))
+        lambda exc: (
+            isinstance(exc, docker.errors.APIError)
+            and any(kw in str(exc).lower() for kw in ("toomanyrequests", "rate limit", "429"))
+        )
     ),
     reraise=True,
     before_sleep=before_sleep_log(logger, logging.WARNING),
--- a/src/cube/backends/local.py
+++ b/src/cube/backends/local.py
@@ -194,16 +194,18 @@ class LocalContainerBackend(ContainerBackend):
         if self.pull_policy == "always" or (self.pull_policy == "missing" and not _image_exists(client, config.image)):
             logger.info("Pulling image %s …", config.image)
             try:
-                self._pull_image(client, config.image)
+                self._pull_image(config.image)
             except docker.errors.APIError as exc:
                 # Convert only after _retry_pull has exhausted all attempts.
                 raise ContainerLaunchError(f"Failed to pull image '{config.image}': {exc}") from exc
         return self._launch_with_retry(config)
 
     @_retry_pull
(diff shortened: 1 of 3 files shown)


amanjaiswal73892:
### Code review

Found 3 issues (all fixed in the follow-up commit [0e51282](https://github.com/The-AI-Alliance/cube-standard/commit/0e51282)):

1. **Incomplete lock coverage** — `cleanup()`, `cleanup_stale()`, and the stale-entry GC path in `list_active()` all called `_save_active()` without acquiring `fcntl.flock`, leaving 3 of 5 write paths to `active.json` unguarded. Under 12+ parallel Ray workers this reliably produced lost-write corruption (reproduced in the regression test: 20 concurrent `_register_active` calls yielded only 4 entries without the lock).

   https://github.com/The-AI-Alliance/cube-standard/blob/c996c03c1be71dd46eb8e2a44695885626794bc1/src/cube/infra_local.py#L653-L658

2. **`os.environ["DOCKER_HOST"]` is a process-global mutation** — the Podman normalisation in `LocalDockerServiceHandle.container` permanently rewrote `DOCKER_HOST` for the entire process lifetime, affecting all subsequent `docker.from_env()` calls in the same Ray worker. Fixed by passing the stripped URL directly to `docker.DockerClient(base_url=...)` instead.

   https://github.com/The-AI-Alliance/cube-standard/blob/c996c03c1be71dd46eb8e2a44695885626794bc1/src/cube/infra_local.py#L285-L290

3. **Stale `DockerClient` reused across retry backoffs** — `launch()` created `client = docker.from_env()` once and passed it into `_pull_image`, which could be retried after waits of up to 5 minutes. If the underlying `requests.Session` timed out during the wait, the retry would fail with a connection error rather than a `429 APIError`, bypassing `_retry_pull`'s predicate and causing an immediate unretried failure — the exact failure mode the PR aimed to prevent. `_launch_with_retry` already created a fresh client per attempt; `_pull_image` now does the same.

   https://github.com/The-AI-Alliance/cube-standard/blob/c996c03c1be71dd46eb8e2a44695885626794bc1/src/cube/backends/local.py#L193-L206

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

amanjaiswal73892 (approved):

PR #130, opened 2026-04-30
fix(benchmark): walk MRO in _is_dynamic instead of any() (#127)

### LINKED ISSUE #127: _is_dynamic() in BenchmarkConfig.__init_subclass__()

NicolasAG:
## Summary
- `_is_dynamic` in `BenchmarkConfig.__init_subclass__` used `any(... for klass in cls.__mro__)`, so a parent's `@property` made the attribute look dynamic even when a child shadowed it with a `ClassVar` — silently skipping the file-load / type-check branch.
- Walk the MRO and return based on the *nearest* definition (parent property + child ClassVar → static, as it should be).
- Adds a regression test that an invalid ClassVar override on a child of a class with `@property benchmark_metadata` is rejected (test fails on the buggy code, passes on the fix).

Closes #127.

> Stacked on #124 (`nico_fix`). Re-target to `main` once #124 lands.

## Test plan
- [x] `uv run pytest tests/test_benchmark.py` — 34 passed
- [x] `make lint` — clean
- [x] Verified the new regression test fails without the fix and passes with it

### DIFF
--- a/src/cube/benchmark.py
+++ b/src/cube/benchmark.py
@@ -329,11 +329,15 @@ class BenchmarkConfig[TTMetadata: TaskMetadata](TypedBaseModel, ABC):
         module_dir = Path(module_file).resolve().parent if module_file else None
 
         def _is_dynamic(name: str) -> bool:
-            """True iff *name* is declared as a ``@property`` somewhere in the MRO
-            below ``BenchmarkConfig`` — meaning the attribute is computed at access
-            time and should not be overwritten with a file-loaded ClassVar value.
+            """True iff the nearest definition of *name* in the MRO is a ``@property``
+            — meaning the attribute is computed at access time and should not be
+            overwritten with a file-loaded ClassVar value. A subclass ClassVar that
+            shadows a parent ``@property`` is treated as static.
             """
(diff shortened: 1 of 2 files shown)


erikchwang (approved):
Thanks for fixing this, LGTM!

PR #133, opened 2026-05-01
feat(benchmark): thread infra through Benchmark.__init__

recursix:
## Why (for Nicolas)

Five cubes today carry the same triple-override of cube-standard's `Benchmark` contract just to thread `infra` from `make(infra)` into `runtime_context`:

| Cube | Status |
|---|---|
| `osworld-cube` | on cube-harness `main` |
| `webarena-verified` | on cube-harness `main` |
| `swebench-verified-cube` | in cube-harness PR #323 |
| `swebench-live-cube` | in cube-harness PR #323 |
| `terminalbench-cube` | in cube-harness PR #323 |

Each one writes:

1. `__init__(self, config, infra=None)` — stash `self._infra`
2. `_setup()` — publish `runtime_context["infra"] = self._infra`
3. `make(infra)` — copy the base's resource-provisioning `for` loop verbatim, just so it can call `MyBenchmark(config=self, infra=resolved_infra)` instead of the base's `benchmark_class(config=self)`

(1) and (3) are pure plumbing — no cube-specific logic. The (3) override copies a loop from base `BenchmarkConfig.make()` and is a maintenance hazard: any future change to base provisioning has to be re-applied five times.

This PR fixes the root cause in cube-standard: `Benchmark.__init__` accepts `infra`, `make()` threads it through. After this lands, the five cubes can drop both `__init__` and `make()` overrides — keeping only their cube-specific `_setup()`. ~140 lines of boilerplate gone in cube-harness, plus the consistency win.

## What changes

Three small edits to `src/cube/benchmark.py`:

1. **`Benchmark.__init__`** — accept `infra: InfraConfig | None = None`, stash `self._infra`.
2. **`BenchmarkConfig.make`** — pass `infra=infra` when constructing `benchmark_class`. Resource-provisioning behaviour is unchanged.
3. **`CompositeBenchmark.__init__`** — accept and forward `infra=None` to match the base signature. The composite itself does not consume it (each sub-benchmark already received its own infra via its own `make(infra)`).

Plus the spec update at `openspec/specs/benchmark/spec.md` and the openspec change folder at `openspec/changes/benchmark-infra-in-init/` (proposal + deltas).

## Backwards compatibility

Fully backwards-compatible. Three test fixtures with custom `__init__(self, config)` were updated to add `infra=None` (`DoneBenchmark`, `FailingTaskConfigsBenchmark`, both in `tests/test_testing.py`). All other `Benchmark` subclasses in this repo inherit the base `__init__` unchanged — they get the new param for free.

For cube-harness:
- Subclasses already declaring `__init__(config, infra=None)` (the five cubes above): signature is now compatible with the base. Override becomes redundant but not breaking — drop in a follow-up.
- Subclasses with `__init__(config)` and no infra (e.g. `miniwob-cube`): need a one-line update to add `infra=None` and forward to `super()`. Or remove their `__init__` if state can move to `_setup()`.

## One subtle behaviour to call out for cube authors

Several existing cubes' `make(infra)` overrides default `infra or LocalInfraConfig()` if no infra is passed. The base `make()` does not apply this default — it forwards whatever the caller passed (including `None`). A cube that drops its `make()` override and relied on this defaulting needs either:
- callers to pass `LocalInfraConfig()` explicitly, **or**
- a small `make()` that resolves the default and calls `super().make(infra=resolved)`.

Documented in proposal.md and benchmark/spec.md.

## Test plan

- [x] `make lint` — clean
- [x] `make test` — 379 passed, 5 skipped (env-specific), 6 deselected
- [x] Two new tests in `tests/test_benchmark.py`: `test_make_threads_infra_to_runtime` and `test_make_without_infra_leaves_runtime_infra_none`

## Follow-ups (not in this PR)

- Cube-harness cleanup PR: drop the `__init__` and `make(infra)` overrides on `osworld-cube`, `webarena-verified`, and (after PR #323 merges) the three SWE cubes. ~140 LOC removed.
- `miniwob-cube` one-line `__init__` update or removal.


NicolasAG:
### Code review

Found 4 issues:

1. `CompositeBenchmarkConfig.make()` constructs `CompositeBenchmark(config=self)` without forwarding `infra`. The PR's own delta spec and the inline comment on `CompositeBenchmark.__init__` say the composite accepts `infra` to match the new base signature, but `make()` never passes it — so `composite._infra` is silently `None` regardless of what the caller passes. The base `BenchmarkConfig.make()` at line 602 was updated correctly; this override was missed.

https://github.com/The-AI-Alliance/cube-standard/blob/385c8fc672b9ae2eb9bf7494da1a3ef346501c84/src/cube/benchmark.py#L965-L975

2. `CompositeBenchmark.__init__` re-introduces `self.config: CompositeBenchmarkConfig = config  # type: ignore[assignment]`, which PR #124 (commit 5e058a4) deliberately removed when `CompositeBenchmark` was parametrised as `Benchmark[\"CompositeBenchmarkConfig\"]`. The generic base already narrows `self.config`, so this re-annotation is redundant and the `# type: ignore` suppresses a check that was clean on `main`.

https://github.com/The-AI-Alliance/cube-standard/blob/385c8fc672b9ae2eb9bf7494da1a3ef346501c84/src/cube/benchmark.py#L768-L776

3. `proposal.md` is `Status: Proposed`. CLAUDE.md's workflow says completed changes get archived under `openspec/changes/archive/YYYY-MM-DD-<name>/` after merge, and PR #121 set the precedent of flipping to `Status: Accepted` on the PR branch (CLAUDE.md says \"For completed changes, move the folder to `openspec/changes/archive/YYYY-MM-DD-<name>/` and apply deltas to the main spec.\").

https://github.com/The-AI-Alliance/cube-standard/blob/385c8fc672b9ae2eb9bf7494da1a3ef346501c84/openspec/changes/benchmark-infra-in-init/proposal.md#L1-L7

4. The migration impact for `Benchmark` subclasses with `__init__(self, config)` (no `infra` kwarg) is a runtime `TypeError` — `benchmark_class(config=self, infra=infra)` will pass an unexpected kwarg. miniwob-cube in cube-harness is one such subclass. proposal.md mentions it but `deltas.md` does not, and the consequence is a hard crash, not a deprecation. Worth surfacing the breakage more prominently in the spec delta.

https://github.com/The-AI-Alliance/cube-standard/blob/385c8fc672b9ae2eb9bf7494da1a3ef346501c84/openspec/changes/benchmark-infra-in-init/proposal.md#L82-L92

Generated with [Claude Code](https://claude.ai/code)

NicolasAG (approved):
Thanks for this. It is much cleaner :)

PR #141, opened 2026-05-05
fix(toolkit): pre-start exec relay via sidecar or python3 at job launch

recursix:
## Problem

`eai job exec` has a ~6% CLOSE_WAIT hang rate per call. The previous exec relay bootstrap required 1–2 `eai job exec` calls at startup (upload script, health-check), meaning every job had a ~12% chance of a multi-minute hang during setup.

## Solution

Start the relay as part of `eai job new`'s launch command — zero bootstrap execs needed.

`relay_startup_args()` emits an if/elif startup script:
1. **Go sidecar** (preferred): if `ToolkitInfraConfig.sidecar_data` is set, the binary is mounted read-only at `/opt/cube-sidecar` via `eai job new --data`. EAI strips the execute bit on data mounts, so the script copies it to `/tmp/cube-sidecar` and `chmod +x`s it before launch.
2. **Python relay** (fallback): if `python3` is available in the image.

By the time the port-forward is established, the relay is already up.

## Changes

- `relay_startup_args()`: extended from python-only to sidecar-preferred if/elif script
- `ToolkitInfraConfig.sidecar_data`: new optional field; when set, adds `--data <name>:/opt/cube-sidecar:ro` to the job launch command
- `_exec_via_relay`: clamp timeout to 86399s; log HTTP error bodies
- `_exec_direct`: 3-attempt CLOSE_WAIT retry loop (eai_timeout = cmd_timeout + 5s, retries=0); graceful error return when all attempts hang
- Remove `_ensure_python3` (dead code — no sudo on Toolkit nodes)

## Test

```
fix-git:       reward=1.000  episode_time=8.3s   (Python relay)
overfull-hbox: reward=1.000  episode_time=17.9s  (Go sidecar — no python3 in LaTeX image)
```

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/container.py
+++ b/cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/container.py
@@ -44,6 +44,8 @@ _EXEC_RELAY_SERVER_PATH = Path(__file__).parent / "_exec_relay_server.py"
 _EXEC_RELAY_PORT = 8787
 _EXEC_RELAY_KICK_TIMEOUT = 15  # eai exec calls during bootstrap; short so CLOSE_WAIT fails fast
 _EXEC_RELAY_HEALTH_TIMEOUT = 15
+# Path inside the container where the sidecar binary is mounted via --data.
+_SIDECAR_MOUNT_PATH = "/opt/cube-sidecar/cube-sidecar"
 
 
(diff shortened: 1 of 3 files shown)


recursix:
Superseded by #142 — that PR includes this commit plus the recoverable-arg-error and assets-mount fixes that the TerminalBench harness depends on.

PR #142, opened 2026-05-05
fix(tool/toolkit): exec-relay sidecar prestart, recoverable arg errors, cube-assets mount

recursix:
## Summary

Three related cube-standard fixes from the TerminalBench-on-toolkit effort, all needed together for the harness to actually drive minimal task images end-to-end. Supersedes #141 (its commit is the first commit on this branch). Also supersedes #140 (fix 2 here is a superset of #140's change).

1. **Pre-start the exec relay at job launch (`cube-infra-toolkit`)** — when a `sidecar_data` blob is mounted at `/opt/cube-sidecar`, run the prebuilt sidecar; otherwise fall back to the inline python3 relay. Eliminates the bootstrap eai-exec round-trips and works on minimal images that lack python3.
2. **Recover from agent argument typos (`cube.tool`)** — a single LLM-side typo like `write_file(path=…, content=…, timeout=120)` previously raised `TypeError`, got wrapped in `StepError`, and `Task.step` terminated the episode. Now the dispatcher pre-binds `action.arguments` against the bound method's signature and returns a recoverable error `Observation` listing the expected parameters, so the agent can correct itself on the next step. Spec updated; 4 new tests cover sync + async unknown-kwarg and missing-required. *(Supersedes #140 which had the same fix without the extracted helper or expected-parameter list.)*
3. **Add `ToolkitInfraConfig.assets_data`** — generalises the `sidecar_data` pattern to a second EAI data slot mounted read-only at `/opt/cube-assets/`. Cubes ship cube-side helper binaries (e.g. a static `uv`) for task images that lack `python3`/`curl`/`apt` sources where the in-container bootstrap can't run. Read-only mount; consumers `cp` + `chmod +x` after copying out.

## Empirical motivation (from TerminalBench run 8 on the 40-task medium subset)

- `StepError`-killed episodes: ≥1/40 → **0/40** after #2.
- Cluster-B evaluators (curl/uvx missing) — `regex-log`, `extract-elf`, `merge-diff-arc-agi-task`, `compile-compcert`, `adaptive-rejection-sampler`: previously all crashed with `passed=0/total=0`; now pytest collects and runs on **39/39 episodes** of the run via the `assets_data` fast path consumed by the cube-harness side (PR #361).

## Test plan

- [x] `pytest tests/test_tool.py` — 39/39 (4 new for the validation paths)
- [x] `make lint` clean
- [x] Live: 8 iterative runs on yul101 toolkit, ~$60 compute. cluster-B fast path activated 39/39 in run 8.
- [ ] Reviewer: try `cube test terminalbench-cube --toolkit --eai-profile yul101 --sidecar-data snow.allac.cube_sidecar`

### DIFF
--- a/cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/container.py
+++ b/cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/container.py
@@ -44,6 +44,8 @@ _EXEC_RELAY_SERVER_PATH = Path(__file__).parent / "_exec_relay_server.py"
 _EXEC_RELAY_PORT = 8787
 _EXEC_RELAY_KICK_TIMEOUT = 15  # eai exec calls during bootstrap; short so CLOSE_WAIT fails fast
 _EXEC_RELAY_HEALTH_TIMEOUT = 15
+# Path inside the container where the sidecar binary is mounted via --data.
+_SIDECAR_MOUNT_PATH = "/opt/cube-sidecar/cube-sidecar"
 
 
 class _ExecRelayUnavailable(Exception):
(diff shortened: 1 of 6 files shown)


cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/container.py
```diff
@@ -44,6 +44,8 @@
 _EXEC_RELAY_PORT = 8787
 _EXEC_RELAY_KICK_TIMEOUT = 15  # eai exec calls during bootstrap; short so CLOSE_WAIT fails fast
 _EXEC_RELAY_HEALTH_TIMEOUT = 15
+# Path inside the container where the sidecar binary is mounted via --data.
+_SIDECAR_MOUNT_PATH = "/opt/cube-sidecar/cube-sidecar"
```
recursix, line 48: The Go sidecar binary is expected at exactly this path. The EAI data mounted via `--data snow.allac.cube_sidecar:/opt/cube-sidecar:ro` must have a file named `cube-sidecar` at its root. The path is kept as a module constant so a future rename only needs to change one place.

cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/container.py
```diff
@@ -155,31 +157,32 @@ def _find_free_port() -> int:
 def relay_startup_args(token: str) -> list[str]:
```
recursix, line 157: The startup script now tries the sidecar binary first (works on any image, no python3 needed), then falls back to the Python relay if python3 is available. Both paths start detached via `nohup ... &` so the job launch command exits quickly. The port-forward connects to the already-running relay — zero `eai job exec` calls on the happy path.

cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/container.py
```diff
@@ -466,7 +431,9 @@ def _exec_via_relay(
         env: Dict[str, str] | None,
     ) -> ExecResult:
         assert self._relay_ready and self._relay_token and self._relay_local_port
-        payload: dict = {"command": command, "timeout": timeout}
+        # Clamp to relay server's 24-hour hard limit (86400s) so we never get a
+        # bad_timeout 400. The real cap is enforced by the caller (e.g. bash tool).
+        payload: dict = {"command": command, "timeout": min(timeout, 86399)}
```
recursix, line 436: The relay server rejects timeouts > 86400s (24h) with a `400 bad_timeout` error. Clamping here prevents that response entirely. Real timeout enforcement is upstream: the `bash` tool caps at 900s, so this 86399s limit will never be reached in practice — it is a safety valve for misconfigured callers.

cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/container.py
```diff
@@ -487,6 +454,19 @@ def _exec_via_relay(
         try:
             with urllib.request.urlopen(req, timeout=timeout + 30) as r:
                 body = json.loads(r.read().decode("utf-8"))
+        except urllib.error.HTTPError as exc:
```
recursix, line 457: The relay server can return HTTP 4xx/5xx instead of a transport error. Previously those fell through to the generic `URLError` handler, losing the response body. Now the body is logged and re-raised as `_ExecRelayUnavailable` so the caller can decide whether to fall back to direct exec.

cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/container.py
```diff
@@ -528,15 +508,31 @@ def _exec_direct(
         logger.info("exec [%s] (direct): %s", self._job_id[:8], command)
         result = None
         duration = 0.0
+        # Use a tight overhead (+5s) so a CLOSE_WAIT hang costs only 5s beyond the
```
recursix, line 511: Tight overhead (+5s instead of +30s) so a CLOSE_WAIT hang on `eai job exec` costs only 5s beyond the actual command timeout, not 30s. `retries=0` prevents `_run_eai` from tripling that overhead — retries are handled by the outer loop so we maintain full control over the retry budget and logging.

cube-resources/cube-infra-toolkit/src/cube_infra_toolkit/toolkit.py
```diff
@@ -54,6 +54,16 @@ class ToolkitInfraConfig(InfraConfig):
     exec_mode: Literal["exec_relay", "direct"] = "exec_relay"
     # Override when eai is not on PATH (e.g. installed in ~/bin via .zshrc).
     eai_path: str = "eai"
+    # EAI data full name to mount as the exec-relay sidecar binary (e.g.
+    # "snow.allac.cube_sidecar").  The data must contain a single file named
+    # "cube-sidecar" at its root; it is mounted read-only at /opt/cube-sidecar.
+    # When set, the relay works on any image — no python3 required.
+    sidecar_data: str | None = None
```
recursix, line 61: Two new optional fields on `ToolkitInfraConfig`: `sidecar_data` (exec-relay Go binary) and `assets_data` (cube-side helpers like `uv`). Both default to `None` so existing recipes need no changes. When set, they add `--data name:/path:ro` to the `eai job new` command, making the blobs available as read-only mounts inside every container this config launches.

src/cube/tool.py
```diff
@@ -251,6 +251,26 @@ def get_action_method(self, action: Action) -> Callable:
             )
         return method
 
+    def _validate_action_args(self, action: Action, method: Callable) -> Observation | None:
```
recursix, line 254: `inspect.signature(method).bind(**arguments)` raises `TypeError` for unknown kwargs and missing required params. We catch it here — before the `try/except` in `execute_action` — so it surfaces as a recoverable `Observation` rather than being wrapped in `StepError` (which terminates the episode). The `params` list in the error message tells the agent exactly which parameter names are valid.

src/cube/tool.py
```diff
@@ -338,6 +358,9 @@ def add(self, a: float, b: float) -> str:
     def execute_action(self, action: Action) -> Observation | StepError:
         """Execute an action by name."""
         method = self.get_action_method(action)
+        invalid = self._validate_action_args(action, method)
```
recursix, line 361: Both sync and async `execute_action` get the pre-validation. The `invalid` check is intentionally before the `try/except` so that argument typos are never logged as exceptions (which would inflate StepError metrics). Pattern is: validate → early-return Observation; execute → catch and wrap in StepError.

PR #146, opened 2026-05-12
fix(cube-infra-azure, cube-computer-tool): retry/backoff + observatio…

kushasareen:
## Summary

- **`cube-infra-azure/azure.py`** — Retry + exponential backoff (with jitter) on `_CachedCliCredential` token fetch. Single-shot credential calls were intermittently failing under `az login` storms during parallel VzM provisioning (n=50 WAA runs); now retries up to 5 times with `2**attempt + random.random()` seconds between attempts before re-raising.
- **`cube-computer-tool/computer.py`** — Hardening for `ComputerBase.execute_action`:
  - Guard observation accumulation against `StepError` so a half-completed action doesn't crash the episode loop.
  - Convert recoverable action exceptions (timeouts, transient network) into `StepError` carrying the tool-call id so trajectories surface the right error context.
  - Handle `None` screenshot from the guest agent (crashed Flask, dead VM) without raising — surface as an evaluator-visible error instead of bringing the worker down.
  - Capture VM state snapshot when a `StepError` is raised so post-hoc diagnostics have something to inspect.
- **`uv.lock`** — Minor lockfile drift.

**Motivation:** these failure modes were the dominant non-agent causes of dropped episodes during the n=50 WAA Azure campaign (~30% of `env_failure` blame in the Sonnet × T2 run, per ch-judge). Both fixes preserve the trajectory's error signal instead of silently truncating it.

### DIFF
--- a/cube-resources/cube-infra-azure/src/cube_infra_azure/azure.py
+++ b/cube-resources/cube-infra-azure/src/cube_infra_azure/azure.py
@@ -78,6 +78,7 @@ from pathlib import Path
 from typing import Any, Literal
 
 from azure.identity import AzureCliCredential
+from azure.identity._exceptions import CredentialUnavailableError as _CredUnavail
 from pydantic import Field, model_validator
 
 from cube.infra_utils import build_volume_setup_script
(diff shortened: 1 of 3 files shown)


amanjaiswal73892 (approved):
LGTM! Essentially brings the same changes as https://github.com/The-AI-Alliance/cube-standard/tree/fix/computer-tool-stepError-on-execute-action

